PR #7, opened 2016-03-08
Extract a "Getting Started Guide"

olivierdagenais:
I moved, combined and improved all the one-time set-up and configuration to `GettingStarted.md`, then linked to it from `README.md` and `Contributing.md`.


GettingStarted.md
```diff
@@ -0,0 +1,181 @@
...
+  1. `Help` -> `Install New Software...`
+  1. In the *Work with* textbox, type `http://archive.eclipse.org/egit/updates-2.1` and click `Add...`
+  1. The *Add Site* dialog appears.  Type `EGit 2.1.0` in the _Name_ textbox and click `OK`.
+  1. Expand both features and select the following plug-ins:
+    * *Eclipse EGit*
+    * *Eclipse EGit - Source*
+    * *Eclipse JGit*
+    * *Eclipse JGit - Source*
+  1. Click `Next >` and finish the wizard to install the plugins
+  1. TODO: install more modern SWT than 4.1 for browser-based authentication
```
jpricketMSFT: Do we want to leave this TODO here? I am not sure what it means myself.

jpricketMSFT:
The changes look good to me. I left one comment about a TODO.

GettingStarted.md
olivierdagenais: > Do we want to leave this TODO here?

Good catch, I had forgotten to resolve that one.  Let me figure this out and I'll replace the TODO with instructions for doing so.

README.md
```diff
@@ -24,51 +24,20 @@ If you prefer to use Git, you can use any Git client with TFS or Team Services a
...
 1. Run ant -Ddir.machine.build-runtime=`<pathToEclipseTarget>`, for example, 
 ```
-(Windows) ant -Ddir.machine.build-runtime=c:\Users\<userId>\dev\eclipseTargets\352\
+(Windows) ant -Ddir.machine.build-runtime=C:\Users\<userId>\dev\eclipseTargets\352\
 (Linux) ant -Ddir.machine.build-runtime=/home/<userId>/dev/eclipseTargets/352/
-(Mac) ant -Ddir.machine.build-runtime=/Applications/eclipse-classic/
+(Mac) ant -Ddir.machine.build-runtime=/Users/<userId>/dev/eclipseTargets/352/
 ``` 
-1. Build results can be found in `build\output`
+1. Build results can be found in `build/output`
```
AlexRukhlin: At this point of reading, the user already "is" in the build folder. So I would say either "... in `output` subfolder of the current directory." or "... in the `build/output` subfolder of the **team-explorer-everywhere** repository."

And the sentence should probably have a period at the end.

GettingStarted.md
```diff
@@ -0,0 +1,181 @@
...
+### Install Java
+
+Two versions of the Java Development Kit are needed: we use Java SE 6 as the minimal supported Java execution environment and to build the release kit, as well as Java SE 8 for running Eclipse.
+
+1. Download and install Java SE 6:
+  * [Windows and Linux: Java Development Kit 6u45](http://www.oracle.com/technetwork/java/javase/downloads/java-archive-downloads-javase6-419409.html#jdk-6u45-oth-JPR)
+  * [Mac: Apple Java 1.6.0_65, a.k.a. Java for OS X 2015-001](https://support.apple.com/kb/DL1572?locale=en_US)
+1. Set the JAVA_HOME environment variable to point to the install, e.g
+ * (Windows) `SET JAVA_HOME=C:\Program Files\Java\jdk1.6.0_45`
+ * (Linux) `JAVA_HOME=~/dev/java/jdk1.6.0_45`
```
AlexRukhlin: Shouldn't we use the -export_ command here? E.g.
export  PATH=$JAVA_HOME/bin:$PATH

GettingStarted.md
```diff
@@ -0,0 +1,181 @@
...
+1. Add JAVA_HOME bin directory to the path
+ * (Windows) `SET PATH=%JAVA_HOME%\bin;%PATH%`
+ * (Linux) `PATH=$JAVA_HOME/bin:$PATH`
+ * (Mac) `PATH=$JAVA_HOME/bin:$PATH`
+1. Download and install Java SE 8 JDK from [Java SE Downloads](http://www.oracle.com/technetwork/java/javase/downloads/index.html).
+
+### Install Ant
+
+1. If you do not already have it, download and install Apache Ant(TM) version 1.9.6 from [Ant Binary Distributions](http://ant.apache.org/bindownload.cgi).
+1. Add the full path of the Ant `bin` directory to the `PATH` system environment variable. You can find more Ant installation details [here](http://ant.apache.org/manual/install.html#installing).
```
AlexRukhlin: I would move "You can find more Ant installation details here." after the command samples.

GettingStarted.md
```diff
@@ -0,0 +1,181 @@
...
+ * (Windows) `SET PATH=C:\dev\apache-ant-1.9.6\bin;%PATH%`
+ * (Linux) `PATH=~/dev/apache-ant-1.9.6/bin:$PATH`
+ * (Mac) `PATH=~/dev/apache-ant-1.9.6/bin:$PATH`
+
+### Install Eclipse
+
+Two versions of Eclipse are needed: we use Eclipse 3.5.2 (Galileo) as the Target Environment, which is the minimum supported Eclipse version (used for building the release and testing backward-compatibility) and Eclipse 4.5.2 (Mars.2) as a dedicated team-explorer-everywhere Development Environment.  Using a dedicated Eclipse installation and workspace for development makes it less disruptive to configure Eclipse-wide preferences.
```
AlexRukhlin: IMO, Eclipse 3.5.2 has nothing with the "Target Environment". We use it as PDE build tool in Ant build scripts.

GettingStarted.md
```diff
@@ -0,0 +1,181 @@
...
+
+1. Download and extract [Eclipse Classic 3.5.2](http://www.eclipse.org/downloads/packages/eclipse-classic-352/galileosr2) to the `dev/eclipseTargets/352` sub-folder under your `HOME` folder.  The absolute path to this folder will be needed later when building with Ant.
```
AlexRukhlin: 1. Why do we first suggest to use the long root path for the Eclipse installation putting the user into troubles with long paths on Windows, and then provide him with the workaround for the artificially created issue? ;-) Looks rather strange for me.
2. Reading the this instruction I feel that I **HAVE** to use `dev/eclipseTarget/352` because it's hardcoded somewhere in TEE code/scripts. May we provide somewhat softer sentence? Say "... _for example_ C:\Eclipse352 or ~/dev/eclipse/352..."?

GettingStarted.md
```diff
@@ -0,0 +1,181 @@
...
+  * (Windows) Use [7-Zip](http://www.7-zip.org/) to [make sure all files are extracted](https://bugs.eclipse.org/bugs/show_bug.cgi?id=166597).
+1. Launch Eclipse 3.5.2
+  1. `Help` -> `Install New Software...`
+  1. In the *Work with* textbox, type `http://archive.eclipse.org/egit/updates-2.1` and click `Add...`
+  1. The *Add Site* dialog appears.  Type `EGit 2.1.0` in the _Name_ textbox and click `OK`.
+  1. Expand both features and select the following plug-ins:
+    * *Eclipse EGit*
+    * *Eclipse EGit - Source*
+    * *Eclipse JGit*
+    * *Eclipse JGit - Source*
```
AlexRukhlin: No need to install source plugins here

GettingStarted.md
```diff
@@ -0,0 +1,181 @@
...
+  1. Click `Next >` and finish the wizard to install the plugins
+  1. TODO: install more modern SWT than 4.1 for browser-based authentication
+1. Download and extract [Eclipse for RCP and RAP Developers 4.5.2](http://www.eclipse.org/downloads/packages/eclipse-rcp-and-rap-developers/mars2)
```
AlexRukhlin: Download and install...

GettingStarted.md
```diff
@@ -0,0 +1,181 @@
...
+  * (Windows) Use [7-Zip](http://www.7-zip.org/) to [make sure all files are extracted](https://bugs.eclipse.org/bugs/show_bug.cgi?id=166597).
```
AlexRukhlin: No need to provide this workaround here as soon as we do not force the user to use long installation root path.

GettingStarted.md
```diff
@@ -0,0 +1,181 @@
...
+ * (Windows) `SET PATH=C:\dev\apache-ant-1.9.6\bin;%PATH%`
+ * (Linux) `PATH=~/dev/apache-ant-1.9.6/bin:$PATH`
+ * (Mac) `PATH=~/dev/apache-ant-1.9.6/bin:$PATH`
+
+### Install Eclipse
+
+Two versions of Eclipse are needed: we use Eclipse 3.5.2 (Galileo) as the Target Environment, which is the minimum supported Eclipse version (used for building the release and testing backward-compatibility) and Eclipse 4.5.2 (Mars.2) as a dedicated team-explorer-everywhere Development Environment.  Using a dedicated Eclipse installation and workspace for development makes it less disruptive to configure Eclipse-wide preferences.
+
+1. Download and extract [Eclipse Classic 3.5.2](http://www.eclipse.org/downloads/packages/eclipse-classic-352/galileosr2) to the `dev/eclipseTargets/352` sub-folder under your `HOME` folder.  The absolute path to this folder will be needed later when building with Ant.
+  * (Windows) Use [7-Zip](http://www.7-zip.org/) to [make sure all files are extracted](https://bugs.eclipse.org/bugs/show_bug.cgi?id=166597).
```
AlexRukhlin: Is that a hard requirement to use 7-zip? Looks like an Ad. ;-)

GettingStarted.md
```diff
@@ -0,0 +1,181 @@
...
+1. Select `Existing local repository`, click the `Add` button.
+  1. Using the `Browse...` button, navigate to your local copy of the **team-explorer-everywhere** repository, select it, and click the `Finish` button.
+1. Click the `Next` button to switch to the Wizard Selection page.
+1. Select the `Import existing Eclipse projects` option and the `source` folder under the root of the repository.
+1. Click the `Next` button.
+1. Make sure that all the `com.microsoft.tfs.*` projects are selected and click the `Finish` button.
+
+### Install Eclipse dependencies
+
+There are additional Eclipse-related dependencies that must be installed _for each Eclipse platform target version_.  For now, let's install the dependencies of earliest-supported Eclipse version (3.5.2).  If you ever need to target another version of Eclipse, repeat this process with the matching version.
```
AlexRukhlin: I would start with the latest target version. And provide a note that the user if he's going to contribute should test against older targets. We also might mention that 3.5 target supposed to be used with TFS on-premises only, but not with VSTS.

GettingStarted.md
```diff
@@ -0,0 +1,181 @@
...
+
+1. Open the `com.microsoft.tfs.client.eclipse.target` project.
+1. Select the `3.5.target` file
```
AlexRukhlin: Period at the end of the sentence.

GettingStarted.md
```diff
@@ -0,0 +1,181 @@
...
+```
+
+## Install development tools
+
+### Install Java
+
+Two versions of the Java Development Kit are needed: we use Java SE 6 as the minimal supported Java execution environment and to build the release kit, as well as Java SE 8 for running Eclipse.
+
+1. Download and install Java SE 6:
+  * [Windows and Linux: Java Development Kit 6u45](http://www.oracle.com/technetwork/java/javase/downloads/java-archive-downloads-javase6-419409.html#jdk-6u45-oth-JPR)
```
AlexRukhlin: Period at the end of the sentence.

GettingStarted.md
```diff
@@ -0,0 +1,181 @@
...
+1. Make sure that all the `com.microsoft.tfs.*` projects are selected and click the `Finish` button.
+
+### Install Eclipse dependencies
+
+There are additional Eclipse-related dependencies that must be installed _for each Eclipse platform target version_.  For now, let's install the dependencies of earliest-supported Eclipse version (3.5.2).  If you ever need to target another version of Eclipse, repeat this process with the matching version.
+
+1. Open the `com.microsoft.tfs.client.eclipse.target` project.
+1. Select the `3.5.target` file
+1. Open the file with the default editor (Target Editor).
+1. Wait until Eclipse downloads indexes of the referenced p2 repositories. (this might take a few minutes)
```
AlexRukhlin: ... (**T**his ...minutes**.**) Note capital "T" and "." before ")"

GettingStarted.md
olivierdagenais: > Shouldn't we use the export command to set environment variables on Linux and Mac?

The thought had crossed my mind, especially since this document is intended to be for one-time instructions and just setting or exporting at the console/terminal won't make it persistent.  Since the intent is to have changes to the environment be persistent, I've replaced the examples with a note to that effect.  If we find enough users/developers tripping up on this part, we can always introduce batch and bash scripts that configure the current console/terminal appropriately, kind of like the "Visual Studio Command Prompt" batch file did.

GettingStarted.md
olivierdagenais: Huh, that's totally my fault: I didn't notice that the 3.5.2 installation wasn't _actually_ used by the 4.5.2 installation when launching (for testing) and that the latter just downloads the parts it needs on demand.  So yes, I conflated two topics together here by assuming the Ant build would also run tests that invoked portions of Eclipse provided by this installation.

To be fair, it was already called "Eclipse Target Environment" before I started shuffling the content around, so I kept the concept.

GettingStarted.md
olivierdagenais: I have an idea to fix this, but it will require a lot more changes to the documentation and I thought I would run it by everybody first:  since we're downloading a number of archives to extract, we could introduce [more explicitly] the concept of a common folder (say, `dev` or `tee`), which lives wherever the user chooses, and under which we suggest everything would go.  It would be a theme throughout the documentation and most paths could be described as relative to that folder.  It would likely contain:

- apache-ant-1.9.6
- eclipse/352
- eclipse/452
- jdk/6
- jdk/8
- team-explorer-everywhere (i.e. the local copy of this Git repo)

What do you think?

GettingStarted.md
olivierdagenais: There's no installer available in the archives, hence why I went with "extract".

GettingStarted.md
olivierdagenais: > the concept of a common folder

This has the potential to simplify Ant builds to not require the `dir.machine.build-runtime` property if we can assume it will be relative to the source tree.  We could even automate the download and extraction of most of the software via a small bootstrap script, although, at the same time, this sort of thing is a solved problem if we use Maven.

GettingStarted.md
olivierdagenais: Hmmm... If we don't point out the potential defect (which may, in fact, occur only in older versions of Windows) and give freedom to the end-user to extract files anywhere (i.e. the "hardcoded" comment earlier), we may be setting them up for failure.  I'd rather they not risk having a broken Eclipse installation if they somehow can't create folders at the root of the system drive and assume they can just extract somewhere else.

GettingStarted.md
olivierdagenais: Would you rather I suggest `jar xvf` [as per comment 11 to bug 166597](https://bugs.eclipse.org/bugs/show_bug.cgi?id=166597#c11)?

GettingStarted.md
olivierdagenais: > 3.5 target supposed to be used with TFS on-premises only, but not with VSTS.

Commit 35348790e9c9db1e7dc41ce4fdfee3f8527be74f fixes that.  Again, I'd rather guide potential contributors into a path of success by pointing to the earliest-supported version of Eclipse, as opposed to a fix or new feature being submitted that only works against the latest version of Eclipse and we don't find out until it's too late.

olivierdagenais:
I think I've addressed all of the feedback.  Thank you for the thorough review and let me know if you'd like to insist on certain items I pushed back.

Contributing.md
```diff
@@ -9,109 +9,11 @@ Interested in contributing to the team-explorer-everywhere project? There are pl
...
-* Click the `Set as Target Platform` link at the top right of the Target Definition view.
-* Open `Window` -> `Preferences` -> Plug-In Development` -> `Target Platform` and 
-make sure the desired target is selected.
-* Clean rebuild projects. (`Project` -> `Clean` -> `Clean all projects`)
+A few styles we follow; most of these are enforced by the Eclipse configuration described in the [Getting Started Guide](GettingStarted.md).
 
-## Code Styles
-A few styles we follow:
-1. No tabs in source code. All tabs should be expanded to 4 spaces.
+1. No tabs in source code. Indentation is done using 4 spaces.
```
AlexRukhlin: ... is to be done ... (?) Sorry my lack of English knowledge.

README.md
```diff
@@ -24,51 +24,20 @@ If you prefer to use Git, you can use any Git client with TFS or Team Services a
...
-git clone https://github.com/Microsoft/team-explorer-everywhere
-```
+Before you can build, please follow the [Getting Started Guide](GettingStarted.md) to install & configure the tools you'll need.
 
-### Build
-Note: The Eclipse target installation location is needed as a parameter for the Ant build variable `dir.machine.build-runtime`. For the samples below, we will assume that the target Eclipse version was installed into '\dev\eclipseTargets\352'.
-1. From a terminal/console window, change to the `build` subfolder of the root folder of the team-explorer-everywhere repository
+Note: The Eclipse target installation location is needed as a parameter for the Ant build variable `dir.machine.build-runtime`. For the samples below, we will assume that the target Eclipse version was installed into the `dev/eclipseTargets/352` sub-folder under your `HOME` folder.
+1. From a terminal/console window, change to the `build` sub-folder of the root folder of the **team-explorer-everywhere** repository.
 1. Run ant -Ddir.machine.build-runtime=`<pathToEclipseTarget>`, for example,
```
AlexRukhlin: Why `pathTo**Target**`? Eclipse 3.5.2 we use in the build is not a target, it's a building tool.

source/com.microsoft.tfs.client.eclipse.target/3.5.target
```diff
@@ -19,5 +19,389 @@
 <unit id="org.slf4j.impl.log4j12.source" version="1.7.2.v20131105-2200"/>
 <repository location="http://download.eclipse.org/tools/orbit/downloads/drops/R20150124073747/repository"/>
 </location>
+<location path="${project_loc}/swt" type="Directory"/>
```
AlexRukhlin: We cannot use SWT from the local disk.

GettingStarted.md
```diff
@@ -0,0 +1,167 @@
...
+1. Download and install Java SE 8 JDK from [Java SE Downloads](http://www.oracle.com/technetwork/java/javase/downloads/index.html).
+
+### Install Ant
+
+1. If you do not already have it, download and extract Apache Ant(TM) version 1.9.6 from [Ant Binary Distributions](http://ant.apache.org/bindownload.cgi).
+1. Ensure the path of the Ant `bin` sub-folder is persistently added to the `PATH` environment variable.
+
+### Install Eclipse
+
+Two versions of Eclipse are needed: we use Eclipse 3.5.2 (Galileo) as the Target Environment, which is the minimum supported Eclipse version (used for building the release) and Eclipse 4.5.2 (Mars.2) as a dedicated team-explorer-everywhere Development Environment.  Using a dedicated Eclipse installation and workspace for development makes it less disruptive to configure Eclipse-wide preferences.
```
AlexRukhlin: We use Eclipse 3.5.2 not as the Target Environment, but as a plugins build tool. The statement about "Target Environment" sounds confusing.

GettingStarted.md
AlexRukhlin: This is a false concept, I'm sorry to say. :-) We should get rid of it. I've explained its history in a new comment on the README.md file.

GettingStarted.md
AlexRukhlin: Personally, I hate to mix sources and tools. Also I prefer may own folders structure on my local disks. It might (and usually does) use some (current) company traditions, which simplifies communication with other developers. Putting Eclipse, Java, Ant into a folder which location is somehow tied to TEE sources looks awkward for me. I use these tools not only for TEE development. I suggested to Jason to use an environment variable pointing to a BuildTools folder, but he preferred to use a command line parameter.

On the other hand, we need that for Ant build only, we expect that all these issues will disappear as soon as we switch to Maven build.

GettingStarted.md
```diff
@@ -0,0 +1,167 @@
...
+
+1. Download and extract [Eclipse Classic 3.5.2](http://www.eclipse.org/downloads/packages/eclipse-classic-352/galileosr2) to the `dev/eclipseTargets/352` sub-folder under your `HOME` folder.  The absolute path to this folder will be needed later when building with Ant.
+  * (Windows) Use [7-Zip](http://www.7-zip.org/) to [make sure all files are extracted](https://bugs.eclipse.org/bugs/show_bug.cgi?id=166597).
```
AlexRukhlin: Again, we should warn the user against using a long root path for the Eclipse instance instead. Note that:
1. The suggested path is not too long yet.
2. The long path issue does nothing to do with Ant build tools preparation. That's a common Eclipse installation issue. That looks strange to see this instruction here.

GettingStarted.md
AlexRukhlin: No, there're many other possibilities. In most cases I use RAR or FAR to unpack archives. Why don't we recommend them? ;-)

PR #10, opened 2016-03-09
User Story #518062 - Allow users to type in URL directly

jpricketMSFT:
Summary of changes:
- Replaced the ServerSelectControl with a simple combo and button.
- Copied the code from the ServerSelectControl to handle the "Servers" button press event
- Copied the code to populate the combo as well
- Made the combo writable (instead of ReadOnly)
- Added AutoComplete for the combo so that the user can see matching elements as they type
- refactored the constructor into several private methods to make the code more readable
- Fixed a bug where the URL for the server wouldn't always show up on the selectionPage

### DIFF
--- a/source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/controls/connect/ServerTypeSelectControl.java
+++ b/source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/controls/connect/ServerTypeSelectControl.java
@@ -5,16 +5,26 @@ package com.microsoft.tfs.client.common.ui.controls.connect;
 
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Set;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
-import org.eclipse.jface.viewers.ISelectionChangedListener;
-import org.eclipse.jface.viewers.SelectionChangedEvent;
+import org.eclipse.jface.fieldassist.ComboContentAdapter;
+import org.eclipse.jface.fieldassist.ContentProposal;
+import org.eclipse.jface.fieldassist.ContentProposalAdapter;
+import org.eclipse.jface.fieldassist.IContentProposal;
+import org.eclipse.jface.fieldassist.IContentProposalProvider;
 import org.eclipse.swt.SWT;
+import org.eclipse.swt.events.ModifyEvent;
+import org.eclipse.swt.events.ModifyListener;
 import org.eclipse.swt.events.SelectionAdapter;
 import org.eclipse.swt.events.SelectionEvent;
 import org.eclipse.swt.layout.GridLayout;
 import org.eclipse.swt.widgets.Button;
+import org.eclipse.swt.widgets.Combo;
 import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Label;
 
@@ -25,25 +35,32 @@ import com.microsoft.tfs.client.common.ui.browser.BrowserFacade.LaunchMode;
 import com.microsoft.tfs.client.common.ui.controls.generic.BaseControl;
 import com.microsoft.tfs.client.common.ui.controls.generic.compatibility.link.CompatibilityLinkControl;
 import com.microsoft.tfs.client.common.ui.controls.generic.compatibility.link.CompatibilityLinkFactory;
+import com.microsoft.tfs.client.common.ui.dialogs.connect.ServerListDialog;
 import com.microsoft.tfs.client.common.ui.framework.command.UICommandExecutorFactory;
 import com.microsoft.tfs.client.common.ui.framework.command.UICommandFinishedCallbackFactory;
 import com.microsoft.tfs.client.common.ui.framework.helper.SWTUtil;
 import com.microsoft.tfs.client.common.ui.framework.layout.GridDataBuilder;
+import com.microsoft.tfs.core.config.persistence.DefaultPersistenceStoreProvider;
 import com.microsoft.tfs.core.util.ServerURIUtils;
 import com.microsoft.tfs.core.util.URIUtils;
+import com.microsoft.tfs.core.util.serverlist.ServerList;
+import com.microsoft.tfs.core.util.serverlist.ServerListConfigurationEntry;
+import com.microsoft.tfs.core.util.serverlist.ServerListManagerFactory;
 import com.microsoft.tfs.util.listeners.SingleListenerFacade;
 
 public class ServerTypeSelectControl extends BaseControl {
     private static final Log log = LogFactory.getLog(ServerTypeSelectControl.class);
 
     private URI serverURI = null;
-    private boolean serverReadonly = false;
     private boolean ignoreServerChangeEvents = false;
     private ICommandExecutor commandExecutor;
     private ICommandExecutor noErrorDialogCommandExecutor;
-    private final ServerSelectControl serverControl;
+    private Combo serverCombo;
+    private ContentProposalAdapter serverComboAdapter;
+    private PropsalProvider proposalProvider;
     private final Button vstsButton;
     private final Button tfsButton;
+    private final Composite serverControl;
     private final SingleListenerFacade listeners = new SingleListenerFacade(ServerTypeSelectionChangedListener.class);
 
     public ServerTypeSelectControl(final Composite parent, final int style) {
@@ -56,18 +73,44 @@ public class ServerTypeSelectControl extends BaseControl {
         layout.verticalSpacing = 0;
         setLayout(layout);
 
-        SelectionAdapter buttonsChanged = new SelectionAdapter() {
+        final SelectionAdapter buttonsChanged = new SelectionAdapter() {
             @Override
             public void widgetSelected(final SelectionEvent e) {
                 onButtonsChanged();
             }
         };
 
-        vstsButton = new Button(this, SWT.RADIO);
-        vstsButton.setText(Messages.getString("ServerTypeSelectControl.BrowseVstsButtonText")); //$NON-NLS-1$
-        vstsButton.setSelection(true);
-        vstsButton.addSelectionListener(buttonsChanged);
-        GridDataBuilder.newInstance().hSpan(layout).hFill().hGrab().applyTo(vstsButton);
+        vstsButton = createTeamServicesButton(layout, buttonsChanged);
+        tfsButton = createTeamFoundationServerButton(layout, buttonsChanged);
+        serverControl = setupServerCombo(layout);
+
+        populateServersCombo();
+
+        if (serverURI != null) {
+            setServerOnCombo(serverURI);
+        } else {
+            setServer(getServerFromCombo());
+        }
+
+        // fire the buttons changed event so that the initial state is updated
+        onButtonsChanged();
+    }
+
+    private Button createTeamFoundationServerButton(final GridLayout layout, final SelectionAdapter buttonsChanged) {
+        final Button button = new Button(this, SWT.RADIO);
+        button.setText(Messages.getString("ServerTypeSelectControl.ConnectButtonText")); //$NON-NLS-1$
+        button.addSelectionListener(buttonsChanged);
+        GridDataBuilder.newInstance().hSpan(layout).hFill().hGrab().vIndent(getVerticalSpacing() * 2).applyTo(button);
+
+        return button;
+    }
+
+    private Button createTeamServicesButton(final GridLayout layout, final SelectionAdapter buttonsChanged) {
+        final Button button = new Button(this, SWT.RADIO);
+        button.setText(Messages.getString("ServerTypeSelectControl.BrowseVstsButtonText")); //$NON-NLS-1$
+        button.setSelection(true);
+        button.addSelectionListener(buttonsChanged);
+        GridDataBuilder.newInstance().hSpan(layout).hFill().hGrab().applyTo(button);
 
         final Label vstsLabel = SWTUtil.createLabel(
             this,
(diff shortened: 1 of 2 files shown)


source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/controls/connect/ServerTypeSelectControl.java
```diff
@@ -180,16 +302,31 @@ public ICommandExecutor getNoErrorDialogCommandExecutor() {
     }
 
     public URI getServer() {
+        if (this.serverURI == null) {
+            return URIUtils.VSTS_ROOT_URL;
```
AlexRukhlin: Why do we need this? Do we use the URL returned in this line?

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/controls/connect/ServerTypeSelectControl.java
```diff
@@ -180,16 +302,31 @@ public ICommandExecutor getNoErrorDialogCommandExecutor() {
...
     }
 
     public void setServer(final URI serverURI) {
+        setServer(serverURI, true);
+    }
+
+    private void setServer(URI serverURI, final boolean setCombo) {
         ignoreServerChangeEvents = true;
-        serverControl.setServerURI(serverURI);
+        if (serverURI != null && URIUtils.VSTS_ROOT_URL.equals(serverURI)) {
```
AlexRukhlin: Do we use URIUtils.VSTS_ROOT_URL only as an indicator of null server URL?

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/controls/connect/ServerTypeSelectControl.java
```diff
@@ -245,4 +382,42 @@ public URI getServerURI() {
             return serverURI;
         }
     }
+
+    private class PropsalProvider implements IContentProposalProvider {
```
AlexRukhlin: Prop**o**salProvider ?

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/controls/connect/ServerTypeSelectControl.java
```diff
@@ -245,4 +382,42 @@ public URI getServerURI() {
...
+        public IContentProposal[] getProposals(final String contents, final int position) {
+            // Work-around to make sure that the popup doesn't open when the
+            // user has opened the dropdown of the combo. By returning an empty
+            // list, the popup won't be shown.
+            if (ignoreServerChangeEvents || serverCombo.getListVisible()) {
+                return new IContentProposal[0];
+            }
+
+            final List<IContentProposal> list = new ArrayList<IContentProposal>(proposals.length);
+            if (contents == null || contents.length() == 0) {
```
AlexRukhlin: StringHelpers.isNullOrEmpty(contents)

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/controls/connect/ServerTypeSelectControl.java
```diff
@@ -245,4 +382,42 @@ public URI getServerURI() {
...
+                for (int i = 0; i < proposals.length; i++) {
+                    list.add(new ContentProposal(proposals[i]));
+                }
+            } else {
+                final String search = contents.toLowerCase().trim();
```
AlexRukhlin: Do we allow spaces in servers? If not we might block then from entering...

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/controls/connect/ServerTypeSelectControl.java
```diff
@@ -245,4 +382,42 @@ public URI getServerURI() {
...
+                for (int i = 0; i < proposals.length; i++) {
+                    if (proposals[i] != null && proposals[i].toLowerCase().contains(search)) {
```
AlexRukhlin: Shouldn't we take care of nulls in the setProposals() method?

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/controls/connect/ServerTypeSelectControl.java
```diff
@@ -245,4 +382,42 @@ public URI getServerURI() {
...
+            // user has opened the dropdown of the combo. By returning an empty
+            // list, the popup won't be shown.
+            if (ignoreServerChangeEvents || serverCombo.getListVisible()) {
+                return new IContentProposal[0];
+            }
+
+            final List<IContentProposal> list = new ArrayList<IContentProposal>(proposals.length);
+            if (contents == null || contents.length() == 0) {
+                for (int i = 0; i < proposals.length; i++) {
+                    list.add(new ContentProposal(proposals[i]));
```
AlexRukhlin: Shouldn't we avoid nulls here as well?

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/controls/connect/ServerTypeSelectControl.java
```diff
@@ -245,4 +382,42 @@ public URI getServerURI() {
             return serverURI;
         }
     }
+
+    private class PropsalProvider implements IContentProposalProvider {
+        private String[] proposals = new String[0];
```
AlexRukhlin: Wouldn't it more convenient to use List<String> here? We might make proposals final.

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/controls/connect/ServerTypeSelectControl.java
```diff
@@ -114,33 +157,112 @@ public void widgetSelected(final SelectionEvent e) {
...
+        final ServerList serverList =
+            ServerListManagerFactory.getServerListProvider(DefaultPersistenceStoreProvider.INSTANCE).getServerList();
+
+        int i = 0;
+        final Set<ServerListConfigurationEntry> servers = serverList.getServers();
+        final String[] list = new String[servers.size()];
+        for (final ServerListConfigurationEntry entry : servers) {
+            list[i++] = entry.getURI().toString();
+        }
+        Arrays.sort(list);
```
AlexRukhlin: Shouldn't we use Collections.sort(list,Collator.getInstance()) here to use locale depending sort order? *list* should be List<String> then.

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/controls/connect/ServerTypeSelectControl.java
```diff
@@ -25,25 +35,32 @@
...
     private boolean ignoreServerChangeEvents = false;
     private ICommandExecutor commandExecutor;
     private ICommandExecutor noErrorDialogCommandExecutor;
-    private final ServerSelectControl serverControl;
+    private Combo serverCombo;
+    private ContentProposalAdapter serverComboAdapter;
+    private PropsalProvider proposalProvider;
     private final Button vstsButton;
     private final Button tfsButton;
+    private final Composite serverControl;
```
AlexRukhlin: Do we need this variable? We don't use it after initialization.

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/controls/connect/ServerTypeSelectControl.java
```diff
@@ -114,33 +157,112 @@ public void widgetSelected(final SelectionEvent e) {
...
-        if (serverURI != null) {
-            serverControl.setServerURI(serverURI);
-        } else if (serverControl.getServerURI() != null) {
-            setServer(serverControl.getServerURI());
-        }
-        serverControl.setEnabled(!serverReadonly);
-        serverControl.addServerSelectionChangedListener(new ISelectionChangedListener() {
+    private Composite setupServerCombo(final GridLayout parentLayout) {
+        final Composite container = new Composite(this, SWT.NULL);
+        final GridLayout layout = SWTUtil.gridLayout(container, 2, false, 0, 0);
```
AlexRukhlin: Why do we need both ..0, 0); and margin assignments below?

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/controls/connect/ServerTypeSelectControl.java
```diff
@@ -25,25 +35,32 @@
...
     private static final Log log = LogFactory.getLog(ServerTypeSelectControl.class);
 
     private URI serverURI = null;
-    private boolean serverReadonly = false;
     private boolean ignoreServerChangeEvents = false;
     private ICommandExecutor commandExecutor;
     private ICommandExecutor noErrorDialogCommandExecutor;
-    private final ServerSelectControl serverControl;
+    private Combo serverCombo;
+    private ContentProposalAdapter serverComboAdapter;
```
AlexRukhlin: It looks like this could be a local variable in the setupServerCombo() method.

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/controls/connect/ServerTypeSelectControl.java
```diff
@@ -114,33 +157,112 @@ public void widgetSelected(final SelectionEvent e) {
...
+        if (server != null) {
+            serverCombo.setText(server.toString());
+        } else {
+            serverCombo.setText(""); //$NON-NLS-1$
+        }
+    }
+
+    private URI getServerFromCombo() {
+        final String server = serverCombo.getText();
+        if (server != null && server.length() > 0) {
```
AlexRukhlin: StringHelpers.isNullOrEmpty(server)

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/controls/connect/ServerTypeSelectControl.java
jpricketMSFT: getServer is called by other classes to get the server the user selected. Null or empty would imply the user didn't select a server (i.e. the Next button should be disabled)

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/controls/connect/ServerTypeSelectControl.java
jpricketMSFT: In the setter, I convert it to null because I want the combo to be cleared. Within this control, null is the same as the VSTS radio button being selected.

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/controls/connect/ServerTypeSelectControl.java
jpricketMSFT: I can never remember that name. I tried to find something under StringUtil and StringUtils. I will fix it.

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/controls/connect/ServerTypeSelectControl.java
```diff
@@ -114,33 +157,112 @@ public void widgetSelected(final SelectionEvent e) {
...
+            return URI.create(server.trim());
```
AlexRukhlin: Can this throw? We often use URIUtils.newURI(server.trim()) in similar cases.

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/controls/connect/ServerTypeSelectControl.java
jpricketMSFT: I could, but what would we do there? remove it? throw an error?

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/controls/connect/ServerTypeSelectControl.java
jpricketMSFT: This method is not called by us. It is called by the SWT control adapter. Therefore, I prefer to check for null. I can use StringHelpers here.

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/controls/connect/ServerTypeSelectControl.java
jpricketMSFT: I don't think URLs are really localizable, but it certainly wouldn't hurt.

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/controls/connect/ServerTypeSelectControl.java
jpricketMSFT: 1) I included it here to keep it the same as the other controls, but it could certainly be a local variable.
2) IMO No. TEE has too many controls. They need to be consolidated. This is the first step towards that. Unless a UI control is going to be shared, it shouldn't be it's own class. If you disagree, we should discuss it. And perhaps I can change this code.

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/controls/connect/ServerTypeSelectControl.java
jpricketMSFT: I copied this code and didn't look at it closely. I will take a closer look.

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/controls/connect/ServerTypeSelectControl.java
jpricketMSFT: I will change it. I wasn't aware of that method.

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/controls/connect/ServerTypeSelectControl.java
```diff
@@ -114,33 +157,112 @@ public void widgetSelected(final SelectionEvent e) {
...
+            serverListCopy);
+
+        populateServersCombo();
+
+        final ServerListConfigurationEntry lastAddedServerListEntry = dialog.getLastAddedServerListEntry();
+
+        if (lastAddedServerListEntry != null) {
+            setServerOnCombo(lastAddedServerListEntry.getURI());
+        } else {
+            setServerOnCombo(lastSelectedServer);
```
AlexRukhlin: 1. Shouldn't we use the server selected in the dialog in this case?
2. lastSelectedServer might be deleted in the dialog.

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/controls/connect/ServerTypeSelectControl.java
```diff
@@ -114,33 +157,112 @@ public void widgetSelected(final SelectionEvent e) {
...
+
+        /*
+         * Note: there is no OK/Cancel on ServerListDialog: all exit paths
+         * should continue.
+         */
+
+        ServerListManagerFactory.getServerListProvider(DefaultPersistenceStoreProvider.INSTANCE).setServerList(
+            serverListCopy);
+
+        populateServersCombo();
```
AlexRukhlin: in populateServersCombo() we unconditionally might set selection and immediately change it in the next statements. Can we do something nice with setting the current selection in serverCombo?

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/controls/connect/ServerTypeSelectControl.java
```diff
@@ -25,25 +35,32 @@
...
 import com.microsoft.tfs.util.listeners.SingleListenerFacade;
 
 public class ServerTypeSelectControl extends BaseControl {
     private static final Log log = LogFactory.getLog(ServerTypeSelectControl.class);
 
     private URI serverURI = null;
-    private boolean serverReadonly = false;
     private boolean ignoreServerChangeEvents = false;
     private ICommandExecutor commandExecutor;
     private ICommandExecutor noErrorDialogCommandExecutor;
```
AlexRukhlin: We do not use these executors. I remember, we already discussed some similar case, but do not remember the decision, I'm sorry to say. :-)

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/controls/connect/ServerTypeSelectControl.java
AlexRukhlin: I believe nowadays they can contain native characters.

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/controls/connect/ServerTypeSelectControl.java
AlexRukhlin: That's my old dream to merge StrigUtil and StringHelpers

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/controls/connect/ServerTypeSelectControl.java
```diff
@@ -245,4 +382,42 @@ public URI getServerURI() {
...
+            // Work-around to make sure that the popup doesn't open when the
+            // user has opened the dropdown of the combo. By returning an empty
+            // list, the popup won't be shown.
+            if (ignoreServerChangeEvents || serverCombo.getListVisible()) {
+                return new IContentProposal[0];
+            }
+
+            final List<IContentProposal> list = new ArrayList<IContentProposal>(proposals.length);
+            if (contents == null || contents.length() == 0) {
+                for (int i = 0; i < proposals.length; i++) {
```
AlexRukhlin: for (final String proposal : proposals)

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/controls/connect/ServerTypeSelectControl.java
AlexRukhlin: for (final String proposal : proposals)

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/controls/connect/ServerTypeSelectControl.java
AlexRukhlin: Ignore

PR #14, opened 2016-03-10
Use only the server part OF VSTS URL in the Add server dialog.

AlexRukhlin:


jpricketMSFT:
The changes look good!

PR #16, opened 2016-03-10
TEE: Technical Debt: Remove all old wizard pages that have been replaced

jpricketMSFT:
Changes for User Story #536241
- Removing several wizard pages that are no longer used
- Removing code that allowed the wizards to show the legacy pages
- Removing controls that are no longer used


AlexRukhlin:
Probably it's good. You know better. :-)

PR #17, opened 2016-03-10
Fixing bug to enable Next button with no existing connection

jpricketMSFT:
I didn't file an issue for this bug. I was bashing on the product today to make sure it was still in good shape and this is the only bug I found. This was a regression caused by one of my changes this week.


AlexRukhlin:
Looks good for me.

PR #19, opened 2016-03-11
Fixing bug with targeting 3.5 - ContentProposal class does not exist

jpricketMSFT:
The ContentProposal implementation of IContentProposal does not exist when targeting Eclipse 3.5. So, I had to implement the interface myself. The only field that we actually use here is the content field itself.


AlexRukhlin:
Looks good for me.

PR #21, opened 2016-03-14
Resize the Folder column when the folder table resizes. 

AlexRukhlin:
image
Before this change the user wasn't able to widen the displayed area of the tree and some long paths were cut.


source/com.microsoft.tfs.client.eclipse.ui.egit/src/com/microsoft/tfs/client/eclipse/ui/egit/importwizard/GitImportWizardSelectFoldersPage.java
```diff
@@ -134,19 +138,50 @@ private void createRepositoriesTreeControl(final Composite container) {
...
         final TreeViewerColumn nameColumnViewer = new TreeViewerColumn(treeViewer, folderColumn);
         nameColumnViewer.setLabelProvider(new FolderNameLabelProvider());
 
-        treeControl.setHeaderVisible(false);
-
         treeViewer.setUseHashlookup(true);
         treeViewer.setContentProvider(new FolderTreeContentProvider());
         treeViewer.setAutoExpandLevel(2);
 
+        container.addControlListener(new ControlAdapter() {
```
jpricketMSFT: I understand why you are doing this, but it seems bad for us to be handling the resize event. SWT should do these kinds of things for us.

jpricketMSFT: If we are doing this in more than one place, we should probably create a helper.

jpricketMSFT:
The changes look good to me.

PR #24, opened 2016-03-15
Ignore serial version UID warnings

AlexRukhlin:
We do not use the standard Java serialization methods, so the serialVersionUID is not needed.


jasholl:
Looks good to me.

PR #26, opened 2016-03-17
Users/jasholl/fix shared tempdir

jasholl:
This was rather tricky to setup, so leaving some notes in case we need to redo this:

The binary jar included here is based on Oli's pull request here:

https://github.com/Microsoft/ApplicationInsights-Java/pull/279

I applied this on top of commit:
66c798e8fe2d4d36058ead2468b46a31428a9854
from https://github.com/Microsoft/ApplicationInsights-Java.git

The tricky part was building the jar.  First I had to set the system level path to point to java16 -- overriding it locally does not work.  Next I ran 'gradlew build' from a windows git bash console -- (straight up cmd window wasn't working because I don't have the various c tools setup to build the dlls).  Then I patched the dlls into the resulting jar with the dlls from the latest ones I found here:

https://azuredownloads.blob.core.windows.net/azuresdkforjava/PackageForAzureLibrariesForJava.html

Hopefully, the next release of the azure libraries will have Oli's pull request.


build/modules/sdk/build.xml
```diff
@@ -90,7 +90,7 @@
                         src="${dir.binaries}/${name.dist-directory.sdk}/redist/lib/log4j-1.2.14.jar" />
             <zipfileset includes="**"
                         excludes="commons-logging.properties META-INF/MSFTSIG.RSA META-INF/MSFTSIG.SF"
-                        src="${dir.binaries}/${name.dist-directory.sdk}/redist/lib/applicationinsights-core-0.9.5.jar" />
+                        src="${dir.binaries}/${name.dist-directory.sdk}/redist/lib/applicationinsights-core-1.0.4-SNAPSHOT.jar" />
```
AlexRukhlin: Shouldn't we rename the JAR removing SNAPSHOT? I doubt that it's OK to ship snapshots. We also might have to take some special care of signing this JAR. Was the original 0.9.5 signed before we included it into the repository?

AlexRukhlin:
I believe, we shouldn't put into the repository unsigned libraries.

build/modules/sdk/build.xml
jasholl: As far as the SNAPSHOT in the name, we should leave it as is, because this is not an official release and is indeed a 'snapshot'.  

I'm not sure about the rules of putting unsigned jars in the repository, but, since this jar is bundled inside another jar that we do sign, we should be good to go at the eclipse level.

I grabbed those dlls from a released version of app insights, so presumably they are signed already.

olivierdagenais:
Obsoleted by #36.

PR #31, opened 2016-03-22
Fix for bug #547695 - Browsing experience asks me to login per account

jpricketMSFT:


jpricketMSFT:
Alex suggested using AtomicReference to return the credentials, so I made those changes.

jpricketMSFT:
Alex said Approved

PR #32, opened 2016-03-22
Bug548495 - TEE: "new" Builds not showing up in the builds page

AlexRukhlin:
The only we can do now is to clarify in TEE labels that they have deas with XAML builds.

### DIFF
--- a/source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/teamexplorer/link/TeamExplorerBaseNavigationLink.java
+++ b/source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/teamexplorer/link/TeamExplorerBaseNavigationLink.java
@@ -27,6 +27,7 @@ public abstract class TeamExplorerBaseNavigationLink implements ITeamExplorerNav
         final TeamExplorerNavigator navigator,
         final TeamExplorerNavigationItemConfig parentNavigationItem) {
 
+        onClick(shell, context, navigator, parentNavigationItem);
     }
 
     protected abstract void onClick(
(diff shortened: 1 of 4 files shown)


DavidStaheli:
Looks good to me

source/com.microsoft.tfs.client.common/src/com/microsoft/tfs/client/common/git/utils/GitHelpers.java
```diff
@@ -30,7 +30,7 @@ public static boolean isEGitInstalled(final boolean activateEGit) {
         } else if (activateEGit) {
             return activateBundle(egit_core) && activateEGitUI();
         } else {
-            return false;
```
jpricketMSFT: It would be good here to have a comment that explains how we got into this else clause and why we should return true.

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/teamexplorer/link/TeamExplorerBaseNavigationLink.java
```diff
@@ -27,6 +27,7 @@ public final void clicked(
         final TeamExplorerNavigator navigator,
         final TeamExplorerNavigationItemConfig parentNavigationItem) {
 
+        onClick(shell, context, navigator, parentNavigationItem);
```
jpricketMSFT: Is this related to the fix?

jpricketMSFT:
Approved with suggestions/questions

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/teamexplorer/link/TeamExplorerBaseNavigationLink.java
AlexRukhlin: It's related to Build navigation items (and some other as well). I put it into a separate commit.

PR #34, opened 2016-03-23
Fixing bug #548507 - Selected Repo not scrolled into view when import…

jpricketMSFT:
Selected Repo not scrolled into view when importing from TEE Repositories page
- added a showSelected call after we set the selection
- also fixed a few missing finals


DavidStaheli:
Looks good

AlexRukhlin:
I'm merging the PR Since it has been approved and that's the last fix we have for the TEE 14.0.3 release.

PR #35, opened 2016-03-23
Improve credentials experience.

AlexRukhlin:
Fixed Bug 550494 - TEE: Sign in page sometimes redirects to the "Create your site URL" page instead of finishing sign in

Fixed few bugs including:
- Bug 550494 - TEE: Sign in page sometimes redirects to the "Create your site URL" page instead of finishing sign in

Got rid of global PATs. There was a bug related to saved global PAT that allowed to use a PAT of one user as a PAT of another one. Having in mind that global PATs are going to retire soon, it would be better not to use them in this release.

The logic of CookieRedentials usage across multiple VSTS accounts has been improved.

Added Credentials and improved PAT validation methods in the TEE REST client.

Some code optimization done in the CredentialsManager. The recently added optional parameter removed from the getCredentials method. It has been used for global PATs only.

The "Next" button's enablement improved on Cross Collection pages.

After all these changes I cannot anymore reproduce the Bug 548491 - TEE: Work Items: 401 trying to create a bug. I hope it's fixed as well.


DavidStaheli:
Looks good to me.  Are any of these changes worth porting to the vso-httpclient-java repo?

AlexRukhlin:
Maybe just the method checkConnection in VssHttpClientBase class. I would suggest to create a user story for that. IMO it might (and should) be a bit improved before we put it into vso-httpclient-java..

PR #40, opened 2016-03-25
Unit test fix.

AlexRukhlin:
It turned out that the test's result depends on JRE version. Fixed.


DavidStaheli:
Looks good.  Thanks!

PR #44, opened 2016-03-28
update app id to show both intellij and eclipse

yacaovsnc:
The branding on this app reads better than the intellij specific one.


jpricketMSFT:
Approved.

PR #45, opened 2016-03-29
Maintain LF-only line endings of clc-launcher through Windows builds

DavidStaheli:
During the process of building the CLC on Windows, it was checking out clc-launcher with core.autocrlf=true.  This replaced LF line endings with CRLF and kept it that way in the release package.  LF-only line endings are necessary for this file, later renamed to 'tf', to be executed on Linux and Mac OS.  This .gitattributes file forces maintenance of LF-only line endings of clc-launcher through Windows builds for subsequent execution on Linux and Mac OS.


jasholl:
Looks good.

PR #51, opened 2016-04-12
Connecting to VSTS root re-use the original connection advisor.

AlexRukhlin:
That ensures that correct proxy settings are used.


jpricketMSFT:
Looks good to me.

DavidStaheli:
Looks good to me too.

PR #53, opened 2016-04-13
Bug 565754 - TEE: NPE in LocalDataAccessLevel

AlexRukhlin:
Bug 565754 - TEE: NPE in the LocalDataAccessLevel.handleClientLocalVersionUpdate method


DavidStaheli:
Looks good!

PR #58, opened 2016-04-15
Bug 567206 - TEE: Improve diagnostic in case of TF400030: The local

AlexRukhlin:
metadata table could not be loaded


DavidStaheli:
Looks good, thanks!

PR #59, opened 2016-04-15
Bug 567174 - CLC prompts for credentials even if /noprompt is specified

AlexRukhlin:
option is specified


jpricketMSFT:
Looks good.

PR #61, opened 2016-04-25
User Story 572790 - TEE: Support Boolean type in query editor

AlexRukhlin:
[User Story 572790](https://mseng.visualstudio.com/DefaultCollection/VSOnline/_workitems/edit/572790)


DavidStaheli:
Looks good, Alex.  Thanks!

PR #62, opened 2016-04-26
Users/yacao/linuxstorage

yacaovsnc:
1. Update authentication library to 0.1.5 which includes enhancement to SWT browsers (we now pass all org.eclipse.swt.browser options to the SWT window, so we can use xulrunner instead of webkit), as well as improvements to gnome-keyring storage.

2. Hook up gnome-keyring secure storage on Linux platform that supports it.  We require gnome-keyring 2.22+ for the simple password API.


source/com.microsoft.tfs.client.clc/src/com/microsoft/tfs/client/clc/prompt/Prompt.java
```diff
@@ -242,17 +242,13 @@ public static UsernamePasswordCredentials getCredentialsInteractively(
...
-                URI.create(REDIRECT_URL),
-                accessTokenStore);
+            authenticator = OAuth2Authenticator.getAuthenticator(CLIENT_ID, REDIRECT_URL, accessTokenStore);
         }
 
         final UserPasswordCredentialProvider provider = new UserPasswordCredentialProvider(authenticator);
 
         try {
-            final Credential tokenCreds = provider.getSpecificCredentialFor(serverURI, PromptBehavior.AUTO, options);
+            final Credential tokenCreds = provider.getCredentialFor(serverURI, PromptBehavior.AUTO, options);
```
yacaovsnc: Just a function name change in auth library.

source/com.microsoft.tfs.core/META-INF/MANIFEST.MF
```diff
@@ -20,8 +20,16 @@ Bundle-ClassPath: com.microsoft.tfs.core.jar,
  libs/jackson-core-2.4.1/jackson-core-2.4.1.jar,
  libs/jackson-core-2.4.1/jackson-databind-2.4.1.jar,
  libs/annotation-detector-3.0.4/annotation-detector-3.0.4.jar,
- libs/java-auth-core/java-auth-core-0.1.0.jar,
- libs/oauth2-useragent/oauth2-useragent-0.7.1.jar
+ libs/jna/jna-4.2.1.jar,
+ libs/jna/jna-platform-4.2.1.jar,
```
yacaovsnc: JNA dependency is required for accessing gnome-keyring.

source/com.microsoft.tfs.core/META-INF/MANIFEST.MF
```diff
@@ -20,8 +20,16 @@ Bundle-ClassPath: com.microsoft.tfs.core.jar,
...
+ libs/oauth2-useragent/oauth2-useragent-0.8.1.jar,
+ libs/slf4j/slf4j-api-1.7.19.jar,
+ libs/slf4j/slf4j-log4j12-1.7.19.jar,
```
yacaovsnc: slf4j dependency is required for the proper logging framework we now use in the auth library.

source/com.microsoft.tfs.core/META-INF/MANIFEST.MF
```diff
@@ -20,8 +20,16 @@ Bundle-ClassPath: com.microsoft.tfs.core.jar,
...
+ libs/azure-interactive-login/aad-interactive-auth-1.0.1-SNAPSHOT.jar,
```
yacaovsnc: Auth library now fall back to MSOpenTech's SWT browser if JavaFx browser won't come up.

source/com.microsoft.tfs.core/META-INF/MANIFEST.MF
```diff
@@ -20,8 +20,16 @@ Bundle-ClassPath: com.microsoft.tfs.core.jar,
...
+ libs/java-auth-lib/java-auth-common-0.1.5.jar,
+ libs/java-auth-lib/java-auth-core-0.1.5.jar,
+ libs/java-auth-lib/java-auth-providers-0.1.5.jar,
+ libs/java-auth-lib/java-auth-storages-0.1.5.jar
```
yacaovsnc: We split the one jar into multiple jars for other projects to consume whatever they need instead of everything.

source/com.microsoft.tfs.core/src/com/microsoft/tfs/core/credentials/CredentialsManagerFactory.java
```diff
@@ -90,6 +91,20 @@ public static CredentialsManager getCredentialsManager(
...
         }
 
+        if (Platform.isCurrentPlatform(Platform.LINUX)) {
+            /*
+             * Linux uses gnome-keyring if it's available unless the user
+             * specifies using PersistanceStoreCredentialsManager
+             */
+            if (usePersistanceCredentialsManager) {
+                return new PersistenceStoreCredentialsManager(persistenceProvider.getConfigurationPersistenceStore());
+            } else {
```
leantk: Might make more sense to make this an "else if" just because else makes me think that the next statement will happen for sure but really its just another if

source/com.microsoft.tfs.core/src/com/microsoft/tfs/core/credentials/internal/GnomeKeyringCredentialsManager.java
```diff
@@ -0,0 +1,133 @@
...
+
+public class GnomeKeyringCredentialsManager implements CredentialsManager {
+
+    private final GnomeKeyringBackedCredentialStore gnomeKeyringStore;
+
+    public static boolean isGnomeKeyringSupported() {
+        return GnomeKeyringBackedCredentialStore.isGnomeKeyringSupported();
+    }
+
+    public GnomeKeyringCredentialsManager() {
```
leantk: NP: It's nice to have the constructor first thing in the class

source/com.microsoft.tfs.core/META-INF/MANIFEST.MF
AlexRukhlin: When used in TEE product, will it use the same log files as the rest of the TEE code does?

source/com.microsoft.tfs.core/src/com/microsoft/tfs/core/credentials/CredentialsManagerFactory.java
AlexRukhlin: I would refactor a bit the code of this method. Too many branches return one and the same new PersistenceStoreCredentialsManager(persistenceProvider.getConfigurationPersistenceStore())

source/com.microsoft.tfs.core/src/com/microsoft/tfs/core/credentials/internal/GnomeKeyringCredentialsManager.java
```diff
@@ -0,0 +1,133 @@
...
+import com.microsoft.tfs.core.util.URIUtils;
+import com.microsoft.tfs.util.Check;
+import com.microsoft.tfs.util.StringUtil;
+
+public class GnomeKeyringCredentialsManager implements CredentialsManager {
+
+    private final GnomeKeyringBackedCredentialStore gnomeKeyringStore;
+
+    public static boolean isGnomeKeyringSupported() {
+        return GnomeKeyringBackedCredentialStore.isGnomeKeyringSupported();
```
AlexRukhlin: How expensive is this call? We do this check twice in a line when creating this credential manager in CredentialsManagerFactory

source/com.microsoft.tfs.core/META-INF/MANIFEST.MF
yacaovsnc: Yes, it's really using log4j and I verified logs went to the same log file.

source/com.microsoft.tfs.core/src/com/microsoft/tfs/core/credentials/internal/GnomeKeyringCredentialsManager.java
```diff
@@ -0,0 +1,133 @@
...
+     * {@inheritDoc}
+     */
+    @Override
+    public boolean removeCredentials(final URI uri) {
+        Check.notNull(uri, "uri"); //$NON-NLS-1$
+
+        // Reduce to scheme, host, port
+        final URI serverURI = URIUtils.removePathAndQueryParts(uri);
+
+        return gnomeKeyringStore.delete(getKey(serverURI));
```
AlexRukhlin: What is the reason not to check gnomeKeyringStore for null here?

source/com.microsoft.tfs.core/src/com/microsoft/tfs/core/credentials/internal/GnomeKeyringCredentialsManager.java
yacaovsnc: The first time is expensive, we need to load the native library libgnome-keyring.so.  The subsequent calls are not as expensive, but we still issue a few calls to gnome-keyring to verify it supports the API we need.

Since we need to make this check anyway, I moved it to a static field as discussed.  This way it's only initialized once during class loading time.

source/com.microsoft.tfs.core/src/com/microsoft/tfs/core/credentials/internal/GnomeKeyringCredentialsManager.java
yacaovsnc: Thanks!  Bugs.. Added the check.

source/com.microsoft.tfs.core/src/com/microsoft/tfs/core/credentials/CredentialsManagerFactory.java
yacaovsnc: Refactored the code, now only creates special secure store when they are available, fall back to `PersistenceStoreCredentialsManager` by default otherwise.

source/com.microsoft.tfs.core/src/com/microsoft/tfs/core/credentials/internal/GnomeKeyringCredentialsManager.java
yacaovsnc: No more constructor.  They aren't interesting.

yacaovsnc:
Addressed all review items.

PR #63, opened 2016-04-28
fix git clone to checkout default branch

jasholl:
the first alphabetical branch was being checked out by default before rather than the default branch.


AlexRukhlin:
Looks good to me

PR #65, opened 2016-05-02
Imporve error message when failed to retrieve VSTS tokens interactively

yacaovsnc:
This is to fix issue 567176 with some caveats:
1. It appears after I introduced the SWT based flow, I can't reliably detect the user cancellation anymore.  This is due to the library don't surface up those errors as the browser is launched in a separate process.  So no cancellation message will be displayed.
2. Now I am catching all exceptions in the auth library so I can try JavaFX first, then try SWT if that doesn't work. Don't need to catch the exceptions anymore.

Added instruction to bypass the browser logins and improved the log message with more details.


DavidStaheli:
Can you please localize the error messages and capitalize Java (instead of "java")?

source/com.microsoft.tfs.client.clc/src/com/microsoft/tfs/client/clc/messages.properties
```diff
@@ -77,7 +77,10 @@ Command.CommandDoesNotSupportOptionFormat=Command ''{0}'' does not support optio
...
-Command.InteractiveAuthenticationFailed=Failed to retrieve credentials interactively.  Check the program log files for details. Interactive authentication is only supported on Oracle Java SE 7 update 6 or higher, Oracle Java SE 8, or OpenJDK 8 runtime.
+Command.InteractiveAuthenticationFailed=Failed to retrieve credentials interactively.  Check the program log files for details. To bypass web browser based interactive authentication, please set the {0} environment variable to true. 
+Command.InteractiveAuthenticationFailedDetailedLog1=Failed to authenticate interactively with web browser. This requires either JavaFX or SWT based web browser control:
+Command.InteractiveAuthenticationFailedDetailedLog2=1. JavaFX web browser control is only supported on Oracle Java SE 7 update 6 or higher, Oracle Java SE 8, or OpenJDK 8 runtime (Please note you may need to compile OpenJFX project yourself).
+Command.InteractiveAuthenticationFailedDetailedLog3=2. To launch SWT browser with a specific XULRunner, please set the value of Java system property 'org.eclipse.swt.browser.XULRunnerPath' to the full path of the specific XULRunner.
```
AlexRukhlin: Do we need to externalize messages we put into the log?

yacaovsnc: At first I didn't localize those strings precisely because they would end up in logs.  Then I thought maybe some users would read logs and they could figure out the problem themselves.  It doesn't hurt to translate those strings.

PR #66, opened 2016-05-05
We do not need to wrap collections on the client side.

AlexRukhlin:
However, we still have to unwrap collections 
when we receive them from the server.


leantk:
LGTM

PR #71, opened 2016-06-03
Consume latest auth library

yacaovsnc:
Latest auth library has many small fixes + gnome keyring improvement.  Also a method signature has been changed so must update that function as well.


DavidStaheli:
Looks good

PR #72, opened 2016-06-06
Bug #599470 - TEE: Null pointer exception in the SCE Properties dialog

AlexRukhlin:
when invoked for the root path "$/"


jeffyoung:
LGTM

PR #76, opened 2016-06-08
Bug #600764 - EE: WorkspaceVersionTable#writeToVersion2 might throw an exception "pendingReconcileCount == this.pendingReconcileCount"

AlexRukhlin:


jasholl:
so, we don't want to actually throw the exception, but instead just log it, right?  If so, looks good to me.

AlexRukhlin:
That's right.

PR #77, opened 2016-06-09
Bug #600362 - TEE: TFS Import wizard (SCE>Import>...) ignores folder selection in SCE

AlexRukhlin:
In case the TFS Import Wizard is invoked from the Source Control Explorer as a context menu action for a folder, the wizard's PageData(ImportOptions) is prepopulated with the folder path that includes the project name as well. We do not need to show the project selection page in this case.

### DIFF
--- a/source/com.microsoft.tfs.client.eclipse.ui/src/com/microsoft/tfs/client/eclipse/ui/wizard/importwizard/TfsImportWizard.java
+++ b/source/com.microsoft.tfs.client.eclipse.ui/src/com/microsoft/tfs/client/eclipse/ui/wizard/importwizard/TfsImportWizard.java
@@ -38,7 +38,9 @@ public class TfsImportWizard extends ImportWizard {
         final IWizardPage nextConnectionPage = getNextConnectionPage();
 
         if (nextConnectionPage != null) {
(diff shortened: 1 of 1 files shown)


source/com.microsoft.tfs.client.eclipse.ui/src/com/microsoft/tfs/client/eclipse/ui/wizard/importwizard/TfsImportWizard.java
```diff
@@ -56,6 +58,11 @@ public IWizardPage getNextPage(final IWizardPage page) {
         return null;
     }
 
+    private boolean skipCrossSelectionPage() {
+        ImportOptions options = (ImportOptions) getPageData(ImportOptions.class);
+        return options.getImportFolders().length > 0;
```
jasholl: null check for getInputFolders() ?

AlexRukhlin: It's never null, i.e. is initialized in the ImportOptions constructor.

PR #79, opened 2016-06-14
Adopt OAuth library

AlexRukhlin:
Use authentication UI provided by OAuth library
if the environment variable TF_USE_OAUTH_LIBRARY is set to TRUE.

OAuth2 token is stored in the in-memory credentials store,
which requires re-authentication to VSTS in each new or long
Eclipse session.

PAT tokens for TFS access are stored in the secure persistant
credentials store, e.g. Windows Credentials Manager or KeyChain.


source/com.microsoft.tfs.client.common/src/com/microsoft/tfs/client/common/credentials/CredentialsHelper.java
```diff
@@ -161,4 +183,103 @@ public static TFSConnection getVstsRootConnection(final TFSConnection connection
...
+
+        /**
+         * {@inheritDoc}
+         */
+        @Override
+        public Token get(String key) {
+            final URI serverURI = URIUtils.newURI(key.split(":", 2)[1]); //$NON-NLS-1$
+            CachedCredentials cachedCredentials = gitCredentialsManager.getCredentials(serverURI);
+            if (cachedCredentials != null) {
+                return new Token(cachedCredentials.getPassword(), TokenType.Access);
```
yacaovsnc: The type should be "Personal".

source/com.microsoft.tfs.client.common/src/com/microsoft/tfs/client/common/credentials/CredentialsHelper.java
```diff
@@ -161,4 +183,103 @@ public static TFSConnection getVstsRootConnection(final TFSConnection connection
...
+            options);
+
+        if (tokenCreds == null) {
+            // OAuth2 token is probably expired. Remove it from the internal
+            // memory store...
+            accessTokenStore.delete("OAuth2:" + URIUtils.VSTS_ROOT_URL_STRING); //$NON-NLS-1$
+
+            // ... and try again.
+            tokenCreds = provider.getCredentialFor(
+                serverURI != null ? serverURI : URIUtils.VSTS_ROOT_URL,
```
yacaovsnc: This call won't work for PAT -- the getCredentialFor() call requires an account URL.  To generate global PATs, we need to call getCredential() which doesn't require a URL parameter.  

The global PAT is also going to be deprecated soon.

source/com.microsoft.tfs.client.common/src/com/microsoft/tfs/client/common/credentials/CredentialsHelper.java
```diff
@@ -161,4 +183,103 @@ public static TFSConnection getVstsRootConnection(final TFSConnection connection
...
+
+        tokenCreds = provider.getCredentialFor(
+            serverURI != null ? serverURI : URIUtils.VSTS_ROOT_URL,
+            PromptBehavior.AUTO,
+            options);
+
+        if (tokenCreds == null) {
+            // OAuth2 token is probably expired. Remove it from the internal
+            // memory store...
+            accessTokenStore.delete("OAuth2:" + URIUtils.VSTS_ROOT_URL_STRING); //$NON-NLS-1$
```
yacaovsnc: It's probably better to call authenticator.signOut() which shields you from internal implementation details.

source/com.microsoft.tfs.client.common/src/com/microsoft/tfs/client/common/credentials/CredentialsHelper.java
```diff
@@ -161,4 +183,103 @@ public static TFSConnection getVstsRootConnection(final TFSConnection connection
...
     }
+
+    public static PatCredentials getOAuthCredentials(final URI serverURI) {
+        final Authenticator authenticator;
+        final Options options = Options.getDefaultOptions();
+        options.patGenerationOptions.displayName = PatCredentials.USERNAME_FOR_CODE_ACCESS_PAT;
+        options.patGenerationOptions.tokenScope = VsoTokenScope.AllScopes;
+
+        if (serverURI != null) {
+            log.debug("Interactively retrieving credential based on oauth2 flow for " + serverURI.toString()); //$NON-NLS-1$
```
yacaovsnc: Those logging and comments were based on "persisting" flag in the tf clc flow.  Maybe we should update it?  I don't think the decision is based on persistence anymore.

source/com.microsoft.tfs.client.common/src/com/microsoft/tfs/client/common/credentials/CredentialsHelper.java
```diff
@@ -161,4 +183,103 @@ public static TFSConnection getVstsRootConnection(final TFSConnection connection
...
+             */
+            authenticator = OAuth2Authenticator.getAuthenticator(CLIENT_ID, REDIRECT_URL, accessTokenStore);
+        }
+
+        final UserPasswordCredentialProvider provider = new UserPasswordCredentialProvider(authenticator);
+
+        Credential tokenCreds;
+
+        tokenCreds = provider.getCredentialFor(
+            serverURI != null ? serverURI : URIUtils.VSTS_ROOT_URL,
```
yacaovsnc: This call won't work for PAT -- the getCredentialFor() call requires an account URL. To generate global PATs, we need to call getCredential() which doesn't require a URL parameter. 

The global PAT is also going to be deprecated soon.

source/com.microsoft.tfs.client.common/src/com/microsoft/tfs/client/common/credentials/CredentialsHelper.java
```diff
@@ -161,4 +174,141 @@ public static TFSConnection getVstsRootConnection(final TFSConnection connection
...
+            authenticator = new VstsPatAuthenticator(CLIENT_ID, REDIRECT_URL, accessTokenStore, tokenStore);
+
+            final String tokenKey =
+                authenticator.getUriToKeyConversion().convert(serverURI, authenticator.getAuthType());
+            removeStalePersonalAccessToken(tokenKey, serverURI);
+
+            token = authenticator.getPersonalAccessToken(
+                serverURI,
+                VsoTokenScope.AllScopes,
+                PatCredentials.USERNAME_FOR_CODE_ACCESS_PAT,
```
yacaovsnc: This is the display name for the PAT on VSTS, do you want to use some other string rather than the hardcoded username?

yacaovsnc:
LGTM.  The only comment is on a display name, feel free to use whatever string you feel is descriptive.

PR #81, opened 2016-06-20
Bug #607651 - TEE: WIT metadata update does not clean up temporary files it creates processing SOAP request.

AlexRukhlin:
STaX processing is included into the standard Java libraries since 1.6.


DavidStaheli:
Looks good to me.

jasholl:
I'm late to the party, but looks good to me too.

PR #82, opened 2016-06-21
Remove duplicate SLF4J binding.

AlexRukhlin:
TEE provides Log4J.jar for binding with Slf4J-API, it seems we do not need to add one more  implementation, i.e. Slf4J-Log4J.jar.


DavidStaheli:
Looks good.  Thanks Alex.

jasholl:
Looks good to me too.

PR #86, opened 2016-07-06
Update TEE to use latest Authentication Library

yacaovsnc:
The latest auth library uses oauth2-useragent and removes dependency on MSOpenTech's AAD auth library.  

Manually tested with Eclipse plugin and CLC, SWT browser login window came up fine, generated PATs and stored in GNOME keyrings on Linux system.


AlexRukhlin:
Looks good for me.
Thanks.

PR #87, opened 2016-07-12
Explicitly specify the "UTF-8" encoding for all SOAP XML processing.

AlexRukhlin:
The main change is in that couple places in SOAP XML processing missed an explicit encoding specification. It looks like after we've switched to a new STaX implementation (the standard JRE one instead of obsolete third party library), the way how the default encoding is determined for XML reader and XML writer have changed.


DavidStaheli:
Looks good, Alex.  Thanks.

AlexRukhlin:
Leave a comment

PR #97, opened 2016-07-18
Rename the prefix when saving PATs in GNOME Keyring

yacaovsnc:
This is because this class is used by both CLC as well as the Eclipse
plugin.  This change means any private TEE 14.0.4 build will need to
re-authenticate after upgrade.  However otherwise this prefix is
incorrect.


AlexRukhlin:
Looks good for me.

PR #98, opened 2016-07-21
Use Java Rest Client library v.15.0.103

AlexRukhlin:
NB! The first commit "Remove old Rest Client" does not have anything interesting but a bulk deletion of the old client sources. You could just skip it especially if you use the Internet Explorer.


source/com.microsoft.tfs.core/rest_core/com/microsoft/alm/client/TeeClientHandler.java
```diff
@@ -0,0 +1,371 @@
...
+                    content = JsonHelper.serializeRequestToString(value);
+                } else {
+                    content = StringUtil.EMPTY;
+                }
+
+                try {
+                    entity = new StringRequestEntity(content, contentType, StringUtil.UTF8_CHARSET);
+                } catch (final UnsupportedEncodingException e) {
+                    // Note that this exception cannot happen at this point
+                    log.error("Illegal content encoding found.", e); //$NON-NLS-1$
```
yacaovsnc: If we are sure this exception can't happen at this point, why not throw it up as a runtime exception?  If we are here something's obviously wrong?

source/com.microsoft.tfs.core/rest_core/com/microsoft/alm/client/TeeClientHandler.java
```diff
@@ -0,0 +1,371 @@
...
+            toRouteDictionary(routeValues, location.getArea(), location.getResourceName());
+
+        final String routeTemplate = location.getRouteTemplate();
+        final String actualTemplate = replaceRouteValues(routeTemplate, dictionary);
+        final URI target = URIUtils.resolve(getBaseUrl(), actualTemplate);
+
+        return URIUtils.addQueryParameters(target, queryParameters);
+    }
+
+    private String replaceRouteValues(final String template, final Map<String, Object> routeValues) {
```
yacaovsnc: P3:  Don't we have a very similar method in the other RestHandler?  Maybe next release we could consolidate this and extract this method to the parent class?

yacaovsnc:
LGTM.

source/com.microsoft.tfs.core/rest_core/com/microsoft/alm/client/TeeClientHandler.java
AlexRukhlin: I do not know why and when (besides of the TFS server/network errors) that could happen. I don't wish to bother the user with this details. IMO, log is the right place for this information.

source/com.microsoft.tfs.core/rest_core/com/microsoft/alm/client/TeeClientHandler.java
AlexRukhlin: No, this method does not exist in the other handler.

source/com.microsoft.tfs.core/rest_core/com/microsoft/alm/client/TeeClientHandler.java
yacaovsnc: That's fine and as long as we can assert that entityMethod.setRequestEntity(null) is okay, and downstream code can handle entity being null, we should be fine. 

If downstream code throws NPE, then the purpose is defeated.

source/com.microsoft.tfs.core/rest_core/com/microsoft/alm/client/TeeClientHandler.java
yacaovsnc: removeUndefinedOptionalParameters() is the name of the other function.  I didn't compare them char by char, but they certainly looks very similar.

PR #105, opened 2016-08-25
Improve OAuth2 behavior:

AlexRukhlin:
Fix result return logic in OAuthCredentialsDialog
    Reduce latency in OAuth2DeviceFlowCallbackDialog
    Improve UX of credential cleanup in ServerListDialog


DavidStaheli:
Looks good.  Thanks, Alex.

source/com.microsoft.tfs.client.common/src/com/microsoft/tfs/client/common/credentials/CredentialsHelper.java
```diff
@@ -224,7 +228,7 @@ private static void removeStaleOAuth2Token() {
         if (oauth2TokenPair != null && oauth2TokenPair.AccessToken != null) {
             final String token = oauth2TokenPair.AccessToken.Value;
 
-            if (!StringUtil.isNullOrEmpty(token) && !isOAuth2TokenValid(token)) {
+            if (force || !StringUtil.isNullOrEmpty(token) && !isOAuth2TokenValid(token)) {
```
jasholl: could you add nested parenthesis to make it more clear?  I believe the order of operations is ( (force || !isNull) && isValid) but seeing the name force makes me think you meant (force || (!isNull && !isValid)

yacaovsnc: Good question.  According to http://docs.oracle.com/javase/tutorial/java/nutsandbolts/operators.html, it appears `&&` take precedence over `||`.

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/dialogs/connect/OAuth2DeviceFlowCallbackDialog.java
```diff
@@ -107,13 +107,17 @@ public void linkActivated(final HyperlinkEvent e) {
...
+         * token polling loop checks the expiration first and then tries to
+         * obtain a token from the identity provider. Thus we set some small
+         * time interval to reach the loop and let it execute once. Since
+         * version 0.4.0 the token polling loop always executes at least one
+         * time, and checks expiration after that. We won't need this interval
+         * after we upgrade to the next library version.
          */
         this.response.getExpiresAt().setTime(new Date());
-        this.response.getExpiresAt().add(Calendar.SECOND, 5);
+        this.response.getExpiresAt().add(Calendar.MILLISECOND, 100);
```
yacaovsnc: I think this is risky.  We need to dispose this UI, and bring the authentication thread back to runnable state, then do the time comparison.  100 ms maybe pushing it on slow machines.  

Maybe we should consider upgrade the auth library instead so we don't have this race condition.

PR #108, opened 2016-09-09
Move protocol handler related calls from Common to Eclipse plugin.

AlexRukhlin:
Move protocol handler related calls from Common to Eclipse plugin.
Improve TFS config. server connection command.


source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/autoconnect/UIAutoConnector.java
```diff
@@ -97,6 +101,9 @@ private void startInternal() {
                 }
 
                 serverURI = UIConnectionPersistence.getInstance().getLastUsedServerURI();
+                if (serverURI != null) {
```
jpricketMSFT: why not log null here?

jpricketMSFT:
LKTM - only one minor comment

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/autoconnect/UIAutoConnector.java
AlexRukhlin: We'll log this situation few lines later. See the line #122

PR #109, opened 2016-09-12
Minor TEE improvements

AlexRukhlin:
- Logging initialization improvement. This allows to collect logging information earlier during the plugin initialization.

- Replace native_registry.dll files with their versions compiled with VS2010. This reduces their size to almost 50%. The DLLs are still not signed though.


jasholl:
looks good to me.

PR #111, opened 2016-09-14
Protocol handler improvements

AlexRukhlin:
#658325 - TEE: Reduce the need to write to HKCR hive by moving MRU
logic to the TEE Configuration files
    #658665 - TEE: Add versioning to the protocol handler launcher script.
    #658797 - TEE: Find a better place to keep the parsed protocol
handler's arguments.


DavidStaheli:
Looks good to me.

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/protocolhandler/ProtocolHandler.java
```diff
@@ -0,0 +1,331 @@
...
+            if (found) {
+                log.info(MessageFormat.format(
+                    "Found the protocol handler argument: {0} {1}", //$NON-NLS-1$
+                    PROTOCOL_HANDLER_ARG,
+                    arg));
+
+                try {
+                    return URIUtils.newURI(arg);
+                } catch (final Exception e) {
+                    log.error("   Incorrect URL in the protocol handler argument", e); //$NON-NLS-1$
```
leantk: NP: Is there supposed to be extra space in front of Incorrect?

leantk:
LGTM too

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/protocolhandler/ProtocolHandler.java
AlexRukhlin: Yes, it is. By idea, in the log it will look like
Found the protocol handler argument ....
       Incorrect URL in the protocol handler argument
It's nice, isn't it? :-)

PR #112, opened 2016-09-21
Task #658615 - TEE: Adopt vsoeclipse protocol with base64 URL encoding.

AlexRukhlin:
Task #658786 - TEE: Improve Git URL parsing to accept empty paths from
on-premises TFS.

### DIFF
--- a/source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/commands/ConnectToDefaultRepositoryCommand.java
+++ b/source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/commands/ConnectToDefaultRepositoryCommand.java
@@ -15,10 +15,13 @@ import com.microsoft.tfs.client.common.framework.command.CommandExecutor;
 import com.microsoft.tfs.client.common.repository.TFSRepository;
 import com.microsoft.tfs.client.common.server.TFSServer;
 import com.microsoft.tfs.client.common.ui.TFSCommonUIClientPlugin;
+import com.microsoft.tfs.client.common.ui.protocolhandler.ProtocolHandler;
+import com.microsoft.tfs.core.TFSConfigurationServer;
 import com.microsoft.tfs.core.TFSConnection;
 import com.microsoft.tfs.core.TFSTeamProjectCollection;
 import com.microsoft.tfs.core.httpclient.Credentials;
 import com.microsoft.tfs.util.Check;
+import com.microsoft.tfs.util.GUID;
 
 /**
  * Connects to the "default" (last used) TFS workspace for a given profile.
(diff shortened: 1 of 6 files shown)


source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/protocolhandler/ProtocolHandler.java
```diff
@@ -23,68 +27,123 @@
...
     }
 
     public boolean hasProtocolHandlerRequest() {
-        return isParsed ? isAvailable : protocolHandlerUri != null;
+        return tryParseProtocolHandlerUri();
+    }
+
+    public String getProtocolHandlerServerUrl() {
+        tryParseProtocolHandlerUri();
+        return isAvailable ? serverUrl.get() : StringUtil.EMPTY;
```
yacaovsnc: Can we make this check without global variables?  The current implementation may be error prone -- if two threads call tryParseProtocolHandlerUri() at the same time, we can't guarantee the isAvailable variable is from the correct thread.

yacaovsnc: Can we make this check without global variables? The current implementation may be error prone -- if two threads call tryParseProtocolHandlerUri() at the same time, we can't guarantee the isAvailable variable is from the correct thread.

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/commands/ConnectToDefaultRepositoryCommand.java
```diff
@@ -80,23 +83,28 @@ protected IStatus doRun(final IProgressMonitor progressMonitor) throws Exception
...
         final SubProgressMonitor projectCollectionMonitor = new SubProgressMonitor(progressMonitor, 1);
 
         try {
-            final GetDefaultProjectCollectionCommand projectCollectionCommand =
-                new GetDefaultProjectCollectionCommand(configurationServer);
-
-            final IStatus projectCollectionStatus =
-                new CommandExecutor(projectCollectionMonitor).execute(projectCollectionCommand);
+            if (ProtocolHandler.getInstance().hasProtocolHandlerRequest()) {
+                GUID collectionId = new GUID(ProtocolHandler.getInstance().getProtocolHandlerCollectionId());
```
yacaovsnc: final

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/protocolhandler/ProtocolHandler.java
```diff
@@ -129,55 +188,50 @@ private URI findProtocolHandlerUriArgument(final String[] applicationArgs) {
...
      * 
-     * -clonefromtfs vsoe://<operation>/?url=<git-repository-url>&Ref=master&EncFormat=UTF8[&<other-parameters>...]
+     * -clonefromtfs vsoeclipse://checkout/?EncFormat=UTF8&tfslink=<base64 encoded parameters>
      * 
-     * At this moment we use vsoi instead of vsoe, and ignore <operation> and
-     * <other-parameters>
      * @formatter:on
      */
-    private synchronized void tryParseProtocolHandlerUri() {
+    private synchronized boolean tryParseProtocolHandlerUri() {
```
yacaovsnc: What if you return a Boolean to indicate the operation is successful or not, and use atomicReference as a parameter to set the isAvailable flag?  Essentially an "out" parameter.

PR #113, opened 2016-09-21
Task #658822 - TEE: Find a better way to generate Web URL in createProtocolHandlerUI.

AlexRukhlin:
Make sure JGit is loaded before we open the Git Repositories page in TEE

Task #658822 - TEE: Find a better way to generate Web URL in createProtocolHandlerUI.


DavidStaheli:
Looks good to me.

PR #120, opened 2016-09-29
Adding the ability to disable telemetry with env var or system property

jpricketMSFT:
These changes are needed to speed up the CLC when being called from an IDE like IntelliJ. It also gives the user a way to turn off telemetry.

### DIFF
--- a/source/com.microsoft.tfs.client.clc/src/com/microsoft/tfs/client/clc/Application.java
+++ b/source/com.microsoft.tfs.client.clc/src/com/microsoft/tfs/client/clc/Application.java
@@ -147,6 +147,10 @@ public abstract class Application implements AbstractConsoleApplication {
          */
         ProductInformation.initialize(ProductName.CLC);
 
+        // We need to check for the disable telemetry flag before we send any
+        // telemetry to the server
+        TfsTelemetryHelper.checkDisableTelemetryProperty();
+
         TfsTelemetryHelper.sendSessionBegins();
         final int ret = run(args, false);
         TfsTelemetryHelper.sendSessionEnds();
@@ -252,7 +256,7 @@ public abstract class Application implements AbstractConsoleApplication {
              * arguments.
              */
 
-            AtomicReference<Exception> outException = new AtomicReference<Exception>();
+            final AtomicReference<Exception> outException = new AtomicReference<Exception>();
             c = parseTokens(args, options, freeArguments, outException);
 
             /*
@@ -485,7 +489,7 @@ public abstract class Application implements AbstractConsoleApplication {
 
                 try {
                     o = optionsMap.findOption(token);
-                } catch (InvalidOptionValueException e) {
+                } catch (final InvalidOptionValueException e) {
                     outException.set(e);
                     break;
                 }
--- a/source/com.microsoft.tfs.client.clc/src/com/microsoft/tfs/client/clc/Application.java
+++ b/source/com.microsoft.tfs.client.clc/src/com/microsoft/tfs/client/clc/Application.java
@@ -41,7 +41,6 @@ import com.microsoft.tfs.core.exceptions.TFSFederatedAuthException;
 import com.microsoft.tfs.core.httpclient.auth.AuthenticationSecurityException;
 import com.microsoft.tfs.core.product.ProductInformation;
 import com.microsoft.tfs.core.product.ProductName;
-import com.microsoft.tfs.core.telemetry.TfsTelemetryHelper;
 import com.microsoft.tfs.core.ws.runtime.exceptions.ProxyException;
 import com.microsoft.tfs.util.Check;
 import com.microsoft.tfs.util.StringUtil;
@@ -147,13 +146,13 @@ public abstract class Application implements AbstractConsoleApplication {
          */
         ProductInformation.initialize(ProductName.CLC);
 
-        // We need to check for the disable telemetry flag before we send any
+        // We need to check for the no telemetry flag before we send any
         // telemetry to the server
-        TfsTelemetryHelper.checkDisableTelemetryProperty();
+        CLCTelemetryHelper.checkNoTelemetryProperty();
 
-        TfsTelemetryHelper.sendSessionBegins();
+        CLCTelemetryHelper.sendSessionBegins();
         final int ret = run(args, false);
-        TfsTelemetryHelper.sendSessionEnds();
+        CLCTelemetryHelper.sendSessionEnds();
 
         return ret;
     }
@@ -340,7 +339,7 @@ public abstract class Application implements AbstractConsoleApplication {
              * Same as above but returned by some low level Core or Common
              * classes, e.g. HttpHost.
              */
-            TfsTelemetryHelper.sendException(e);
+            CLCTelemetryHelper.sendException(e);
 
             final String messageFormat = Messages.getString("Application.AnArgumentErrorOccurredFormat"); //$NON-NLS-1$
             final String message = MessageFormat.format(messageFormat, e.getLocalizedMessage());
@@ -406,7 +405,7 @@ public abstract class Application implements AbstractConsoleApplication {
              * The most basic core exception class. All lower level (SOAP)
              * exceptions are wrapped in these.
              */
-            TfsTelemetryHelper.sendException(e);
+            CLCTelemetryHelper.sendException(e);
 
             final String messageFormat = Messages.getString("Application.AnErrorOccurredFormat"); //$NON-NLS-1$
             final String message = MessageFormat.format(messageFormat, e.getLocalizedMessage());
@@ -414,7 +413,7 @@ public abstract class Application implements AbstractConsoleApplication {
             display.printErrorLine(message);
             ret = ExitCode.FAILURE;
         } catch (final Throwable e) {
-            TfsTelemetryHelper.sendException(new Exception("Unexpected exception.", e)); //$NON-NLS-1$
+            CLCTelemetryHelper.sendException(new Exception("Unexpected exception.", e)); //$NON-NLS-1$
 
             log.error("Unexpected exception: ", e); //$NON-NLS-1$
         }
--- a/source/com.microsoft.tfs.client.clc/src/com/microsoft/tfs/client/clc/telemetry/CLCTelemetryHelper.java
+++ b/source/com.microsoft.tfs.client.clc/src/com/microsoft/tfs/client/clc/telemetry/CLCTelemetryHelper.java
@@ -14,6 +14,11 @@ import com.microsoft.tfs.core.telemetry.TfsTelemetryHelper;
 
 public class CLCTelemetryHelper extends TfsTelemetryHelper {
     public static void sendCommandFinishedEvent(final Command command, final int retCode) {
+        if (sendingTelemetryDisabled) {
+            // Don't send any telemetry
+            return;
+        }
+
         final String commandName;
         if (command == null) {
             commandName = "**UNKNOWN**"; //$NON-NLS-1$
--- a/source/com.microsoft.tfs.client.clc/src/com/microsoft/tfs/client/clc/telemetry/CLCTelemetryHelper.java
+++ b/source/com.microsoft.tfs.client.clc/src/com/microsoft/tfs/client/clc/telemetry/CLCTelemetryHelper.java
@@ -7,14 +7,31 @@ import java.text.MessageFormat;
 import java.util.HashMap;
 import java.util.Map;
 
+import com.microsoft.tfs.client.clc.EnvironmentVariables;
 import com.microsoft.tfs.client.clc.ExitCode;
 import com.microsoft.tfs.client.clc.commands.Command;
 import com.microsoft.tfs.core.telemetry.TfsTelemetryConstants;
 import com.microsoft.tfs.core.telemetry.TfsTelemetryHelper;
 
 public class CLCTelemetryHelper extends TfsTelemetryHelper {
+    // A property to define whether collecting telemetry should be skipped.
+    private static final String NO_TELEMETRY_PROPERTY_NAME = "com.microsoft.tfs.client.clc.telemetry.notelemetry"; //$NON-NLS-1$
+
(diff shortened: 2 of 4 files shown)


source/com.microsoft.tfs.core/src/com/microsoft/tfs/core/telemetry/TfsTelemetryHelper.java
```diff
@@ -75,22 +81,42 @@ public synchronized static TelemetryClient getTelemetryClient() {
     }
 
     public static void sendMetric(final String name, final double value) {
+        if (sendingTelemetryDisabled) {
```
yacaovsnc: It appears [TelemetryConfiguration ](https://github.com/Microsoft/ApplicationInsights-Java/blob/master/core/src/main/java/com/microsoft/applicationinsights/TelemetryConfiguration.java)has a "disable tracking" configuration, maybe we can leverage that instead?

Its method to disable tracking:
```
    public void setTrackingIsDisabled(boolean disable) {
        trackingIsDisabled = disable;
    }
```

yacaovsnc: Another random thought...  If TelemetryClient is an interface instead of an implementation, it would be so much easier to do this - we could just return a no-op client from the getTelemetryClient() function and put all this logic in a single place.  However since TelemetryClient is an implementation, we are forced to short circuit the logic in many places.  I think this example really show case the "refactorablity" of programming against interface vs programming against implementation...  Oh well...

source/com.microsoft.tfs.client.clc/src/com/microsoft/tfs/client/clc/Application.java
```diff
@@ -147,6 +147,15 @@ public int run(final String[] args) {
...
 
+        // We need to check for the -notelemetry option before we send any
+        // telemetry to the server
+        for (final String arg : args) {
+            if (arg.equalsIgnoreCase("-notelemetry") || //$NON-NLS-1$
+                arg.equalsIgnoreCase("/notelemetry")) { //$NON-NLS-1$
+                TfsTelemetryHelper.disableSendingTelemetry();
+            }
+        }
+
```
AlexRukhlin: May we use an environment variable instead? E.g.
        if (EnvironmentVariables.getBoolean(EnvironmentVariables.DISABLE_TELEMETRY, false)) {
            TfsTelemetryHelper.disableSendingTelemetry();
        }

source/com.microsoft.tfs.client.clc/src/com/microsoft/tfs/client/clc/vc/Main.java
```diff
@@ -52,7 +55,29 @@ public void uncaughtException(final Thread t, final Throwable e) {
...
 
         try {
-            ret = main.run(args);
+            String[] commandArgs = args;
+
+            // Look for arg==@, this means that the user wants to type in the
+            // arguments one line at a time
+            // This was implemented to allow other programs to spin up the CLC
+            // and then pass arguments via stdin
+            if (args.length == 1 && args[0].equals("@")) { //$NON-NLS-1$
```
AlexRukhlin: Why not use the existing functionality, i.e. the Application#runCommandFile method?

source/com.microsoft.tfs.core/src/com/microsoft/tfs/core/telemetry/TfsTelemetryHelper.java
jpricketMSFT: In this case I want to avoid creating the configuration instance and the client instance to save as much time as possible. The more code I can skip the more time I save. But I didn't notice that method. I may try that method and profile with it disabled to see if that provides enough of a savings.

yacaovsnc: Just playing the devil's advocate:  would >0.5 seconds make any difference if we have left the initialization code as a static initializer?  I assume the jvm would take a while to startup, and can mask the time it takes to initialize this class?  So if we are going to mask the JVM start time, we probably should start the JVM before user tries to do their first action; and if we start JVM in advance, we can probably ignore this initialization time?    

On the other hand, the disabling logic is spread in quite a few places (and even in multiple classes as your modification shows), it's really not DRY.

AlexRukhlin: We might create
```
    private static class DummyTelemetryClient extends TelemetryClient {
        @Override
        public boolean isDisabled() {
            return true;
        }

        @Override
        public void trackEvent(String name, Map<String, String> properties, Map<String, Double> metrics) {
            return;
        }

        @Override
        public void trackException(Exception exception) {
            return;
        }

        @Override
        public void trackSessionState(SessionState sessionState) {
            return;
        }

        @Override
        public void trackPageView(PageViewTelemetry telemetry) {
            return;
        }

        @Override
        public void trackPageView(String name) {
            return;
        }

        @Override
        public void trackMetric(String name, double value) {
            return;
        }
    }
```
and instantiate it in
```
    public synchronized static TelemetryClient getTelemetryClient() {
        if (aiClient == null) {
            if (sendingTelemetryDisabled) {
                aiClient = new DummyTelemetryClient();
            } else {
               . . .
            }
        }

        return aiClient;
    }
```

AlexRukhlin: Or maybe even simpler:
```
    public synchronized static void disableSendingTelemetry() {
        aiClient = new DummyTelemetryClient();
    }
```

yacaovsnc: I agree no-op clients are the best way to be safe, this would shield us from bugs in TelemetryClient as well.  Just want to point out since TelemetryClient isn't an interface, it would expose (in the future) new public classes and our dummy may suddenly gain new abilities.  However if all our access to this client is again encapsulated in our own helpers, the risk is pretty low IMHO.

jpricketMSFT: I tried this as well and profiled the CLC. Unfortunately just calling the constructor for the TelemetryClient (even from a derived class like this one) causes the TelemetryConfiguration class to get initialized and causes the time to be wasted. It seems the only way to avoid this time penalty is to put the checks in the classes that I already have changed.

source/com.microsoft.tfs.client.clc/src/com/microsoft/tfs/client/clc/Application.java
jpricketMSFT: A command line option (even an undocumented one) seemed like a better approach. Do we use environment variables in other places like this?

source/com.microsoft.tfs.client.clc/src/com/microsoft/tfs/client/clc/vc/Main.java
jpricketMSFT: This worked perfectly :) - I didn't have to very much at all

jpricketMSFT: This worked perfectly :) - I didn't have to change anything in the CLC, just some of the code in IntelliJ

jpricketMSFT:
I removed the Option as well as the code to do the @ stuff.

source/com.microsoft.tfs.core/src/com/microsoft/tfs/core/telemetry/TfsTelemetryHelper.java
```diff
@@ -44,8 +50,21 @@
...
 
+    public synchronized static void checkDisableTelemetryProperty() {
+        // Unfortunately, we can't push the disabled logic onto the
+        // TelemetryClient itself because the initialization code for
+        // that object simply takes too much time. So, we have to stop
+        // the sending of telemetry here in the helper class.
+
+        // If the system property exists or the env var is true, then we disable
+        // telemetry.
+        sendingTelemetryDisabled = System.getProperty(DISABLE_TELEMETRY_PROPERTY_NAME) != null
```
yacaovsnc: Would a "-Dcom.microsoft.tfs.core.telemetry.disabletelemetry=false" still disable it anyway?  I guess we are using this as a flag instead of a property?

jpricketMSFT: Correct. I am using it as a flag. If this is not ideal, I can remove the System Property. I ended up using the env var from IntelliJ - it was just easier. I kept the System property because I thought we might end up going that route if we need to later.

yacaovsnc: This is fine as I don't see an easy way to read flags -- I was thinking if there is a way to pass the flag in with '-X'  instead of '-D', that would make the distinction clear.  However I couldn't find any way to read '-X' values, looks like everything is specific to oracle settings.  So nothing we can do!

source/com.microsoft.tfs.client.clc/src/com/microsoft/tfs/client/clc/Application.java
```diff
@@ -147,6 +147,10 @@ public int run(final String[] args) {
          */
         ProductInformation.initialize(ProductName.CLC);
 
+        // We need to check for the disable telemetry flag before we send any
+        // telemetry to the server
+        TfsTelemetryHelper.checkDisableTelemetryProperty();
```
AlexRukhlin: Shouldn't we use the CLCTelemetryHelper class here and in two other calls in this method? (For consistency :-))

source/com.microsoft.tfs.client.clc/src/com/microsoft/tfs/client/clc/telemetry/CLCTelemetryHelper.java
```diff
@@ -14,6 +14,11 @@
 
 public class CLCTelemetryHelper extends TfsTelemetryHelper {
     public static void sendCommandFinishedEvent(final Command command, final int retCode) {
+        if (sendingTelemetryDisabled) {
```
AlexRukhlin: I would create a getter isSendingTelemetryDisabled()

AlexRukhlin: I would create a getter isSendingTelemetryDisabled()

source/com.microsoft.tfs.client.clc/src/com/microsoft/tfs/client/clc/Application.java
AlexRukhlin: Shouldn't we use the CLCTelemetryHelper class here and in two other calls in this method? (For consistency :-))

source/com.microsoft.tfs.core/src/com/microsoft/tfs/core/telemetry/TfsTelemetryHelper.java
```diff
@@ -29,8 +30,13 @@
     private static final Log log = LogFactory.getLog(TfsTelemetryHelper.class);
 
     private static TelemetryClient aiClient;
+    protected static boolean sendingTelemetryDisabled = false;
 
-    static {
+    // A property to define whether collecting telemetry should be skipped.
+    private static final String DISABLE_TELEMETRY_PROPERTY_NAME = "com.microsoft.tfs.core.telemetry.disabletelemetry"; //$NON-NLS-1$
+    private static final String DISABLE_TELEMETRY_ENVIRONMENT_VAR = "disabletelemetry"; //$NON-NLS-1$
```
AlexRukhlin: May we move this constant to the EnvironmentVariables class?

AlexRukhlin: May we move this constant to the EnvironmentVariables class?
To be consistent, may we define the constant as DISABLE_TELEMETRY = "TF_DISABLE_TELEMETRY"

source/com.microsoft.tfs.core/src/com/microsoft/tfs/core/telemetry/TfsTelemetryHelper.java
AlexRukhlin: Can we move this initialization to the CLCTelemetryHelper class? We might make it just as a static initializer. e.g.
    static {
        sendingTelemetryDisabled = System.getProperty(DISABLE_TELEMETRY_PROPERTY_NAME) != null
            || EnvironmentVariables.getBoolean(DISABLE_TELEMETRY_ENVIRONMENT_VAR, false);
    }

AlexRukhlin: One more thing. As we wish to disable telemetry only in the specific case of the TF command usage (IntelliJ plugin), Doesn't make sense to create in the base TfsTelemetryHelper class a protected getter isTelemetryDisabled() {return false;}, and override it in the CLCTelemetryHelper class?

source/com.microsoft.tfs.client.clc/src/com/microsoft/tfs/client/clc/Application.java
AlexRukhlin: Actually, there are more places in the Application class where TfsTelemetryHelper is used. We should use CLCTelemetryHelper in all of them.

source/com.microsoft.tfs.core/src/com/microsoft/tfs/core/telemetry/TfsTelemetryHelper.java
jpricketMSFT: I moved all the logic of reading the environment and setting the flag to the CLC helper, but I can't override methods because the classes are static classes :(

PR #124, opened 2016-10-11
Adding owner-display-name to the xml format of the Workspaces command

jpricketMSFT:
This change to allow IntelliJ to display the owner of the workspace properly. The owner name which is currently available seems to come back as a guid sometimes.


DavidStaheli:
Looks good to me.

PR #129, opened 2016-10-17
Task #688521 - Fix PAT refresh method

AlexRukhlin:


source/com.microsoft.tfs.client.common/src/com/microsoft/tfs/client/common/credentials/CredentialsHelper.java
```diff
@@ -51,115 +47,53 @@
...
-            final URI baseURI = connection.getBaseURI();
-            final CachedCredentials cachedCredentials = gitCredentialsManager.getCredentials(baseURI);
-
-            if (cachedCredentials != null && cachedCredentials.isPatCredentials()) {
-                final PatCredentials patCredentials = (PatCredentials) cachedCredentials.toCredentials();
-                return isCredentialsValid(connection, PreemptiveUsernamePasswordCredentials.newFrom(patCredentials));
-            }
+        if (cachedCredentials.isPatCredentials() == currentCredentials.isPatCredentials()) {
+            // The credentials are changed and are of the same type, i.e either
+            // both PAT, or both Alternative, or both Basic. Let's refresh the
```
jpricketMSFT: the comment is misleading. It implies there are 3 types of creds (PAT, Alternate, Basic), but the if statement implies only 2 (PAT and not PAT). Perhaps the comment can be changed to make it clear?

PR #131, opened 2016-11-07
Continue querying different collections if one instance failed.

yacaovsnc:


jpricketMSFT:
Looks good to me as well

source/com.microsoft.tfs.client.eclipse.ui.egit/src/com/microsoft/tfs/client/eclipse/ui/egit/importwizard/WizardCrossCollectionRepoSelectionPage.java
```diff
@@ -138,7 +138,18 @@ protected void clearList() {
 
     @Override
     protected void appendCollectionInformation(final TFSTeamProjectCollection collection) {
-        final VersionControlClient vcClient = collection.getVersionControlClient();
+        final VersionControlClient vcClient;
+        try {
+            vcClient = collection.getVersionControlClient();
```
yacaovsnc: Alex stopped by and we chatted.  It appears this is the only place it could throw, and if we catch at the caller level, we risk skipping too much.  

This works for the current issue, as from the log it indeed appears we are throwing when trying to get a VC client.

PR #136, opened 2016-11-09
Bug #706718 - No file information provided in case renaming of a file is impossible.

AlexRukhlin:


source/com.microsoft.tfs.util/src/com/microsoft/tfs/util/temp/TempStorageService.java
```diff
@@ -343,8 +344,19 @@ public synchronized void forgetItem(final File item) {
...
         Check.notNull(sourceItem, "sourceItem"); //$NON-NLS-1$
         Check.notNull(targetItem, "targetItem"); //$NON-NLS-1$
-        Check.isTrue(sourceItem.exists(), "renaming: source file does not exist"); //$NON-NLS-1$
-        Check.isTrue(!targetItem.exists(), "renaming: target file already exists"); //$NON-NLS-1$
+
+        if (!sourceItem.exists()) {
+            throw new RuntimeException(
+                MessageFormat.format(
+                    Messages.getString("TempStorageService.RenameErrorSourceDoesNotExistFormat"), //$NON-NLS-1$
+                    sourceItem.getAbsolutePath()));
```
madhurig: The message takes 2 params, but only sourceItem is passed. Shouldn't the targetItem to which the rename failed be supplied here too?

source/com.microsoft.tfs.util/src/com/microsoft/tfs/util/temp/TempStorageService.java
```diff
@@ -343,8 +344,19 @@ public synchronized void forgetItem(final File item) {
...
+        }
+        if (targetItem.exists()) {
+            throw new RuntimeException(
+                MessageFormat.format(
+                    Messages.getString("TempStorageService.RenameErrorTargetExistsFormat"), //$NON-NLS-1$
```
madhurig: Same as above, message string has {0} and {1} params but only one source item is passed.

source/com.microsoft.tfs.util/src/com/microsoft/tfs/util/temp/TempStorageService.java
AlexRukhlin: Thanks!

PR #147, opened 2016-12-13
Fixing NullReferenceException error when doing KeepYours

jpricketMSFT:
Fixing NullReferenceException error when doing KeepYours and the original file was deleted on the server.

### DIFF
--- a/source/com.microsoft.tfs.client.clc/src/com/microsoft/tfs/client/clc/vc/commands/CommandResolve.java
+++ b/source/com.microsoft.tfs.client.clc/src/com/microsoft/tfs/client/clc/vc/commands/CommandResolve.java
@@ -12,6 +12,7 @@ import java.util.List;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicReference;
 
+import com.microsoft.alm.client.utils.StringUtil;
 import com.microsoft.tfs.client.clc.AcceptedOptionSet;
 import com.microsoft.tfs.client.clc.EnvironmentVariables;
 import com.microsoft.tfs.client.clc.ExitCode;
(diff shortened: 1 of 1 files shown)


source/com.microsoft.tfs.client.clc/src/com/microsoft/tfs/client/clc/vc/commands/CommandResolve.java
```diff
@@ -203,7 +203,9 @@ public void run() throws ArgumentException, MalformedURLException, CLCException,
          */
         boolean containsSymlink = false;
         for (final Conflict conflict : conflicts) {
-            if (FileSystemUtils.getInstance().getAttributes(conflict.getTargetLocalItem()).isSymbolicLink()) {
+            if (conflict.getTargetLocalItem() != null
+                && conflict.getTargetLocalItem().length() > 0
```
AlexRukhlin: Can we use !StringUtil.isNullOrEmpty(conflict.getTargetLocalItem()) instead of two checks above?

AlexRukhlin: Doesn't the third option || resolution.equals(Resolution.OVERWRITE_LOCAL looks strange here? The message we display says: "Symbolic link conflicts can only be resolved by keeping the local version (KeepYours) or taking the server version 
 (TakeTheirs)."

jpricketMSFT: I agree, but I would hesitate to change the code since I don't know the best way to test it.

PR #151, opened 2016-12-15
intl update

kgliang:
add localiation md and images, updated the readme.md


img/Localization.md
```diff
@@ -0,0 +1,24 @@
+We’re using an open localization platform to regularly integrate community feedback and contributions into our products.
```
AlexRukhlin: Do we need this file?

PR #161, opened 2017-01-19
Use latest auth library that handles tenant

yacaovsnc:
Auth library is updated with one more function to take in an OAuth2 token pair when generating PAT.

In TEE, there are two code paths that lead to calling the CredentialHelper to generate a PAT:

1. Connect directly to a server account url.  In this case, we do not have an oauth2 token stored, we call the auth library with no OAuth2Token and an valid account url.  The auth library will detect tenant and prompt the user to login with the proper tenant, and with the oauth2 token, create a PAT.

2. Browsing and create a configuration for all accounts.  In this case, we first call the auth library to get an OAuth2 token, and we use "common" placeholder in order to get that.  Afterwards, we pass in this oauth2 token to auth library with each specific account url.  Auth library will use the oauth2 token to create PAT, and not prompt the user.


source/com.microsoft.tfs.client.common/src/com/microsoft/tfs/client/common/credentials/CredentialsHelper.java
```diff
@@ -79,24 +79,29 @@ public static void refreshCredentialsForGit(final TFSConnection connection) {
             gitCredentialsManager.setCredentials(currentCredentials);
```
AlexRukhlin: return;

source/com.microsoft.tfs.client.common/src/com/microsoft/tfs/client/common/credentials/CredentialsHelper.java
```diff
@@ -79,24 +79,29 @@ public static void refreshCredentialsForGit(final TFSConnection connection) {
...
         }
 
-        if (cachedCredentials.equals(currentCredentials)) {
-            // The credentials haven't changed.
-            // No need to refresh.
-            return;
-        }
+        if (cachedCredentials != null) {
```
AlexRukhlin: No need to check for null here. We should return earlier if cachedCredentials == null

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/wizard/common/WizardServerSelectionPage.java
```diff
@@ -130,7 +134,9 @@ protected boolean onPageFinished() {
         }
     }
 
-    private List<Account> getUserAccounts(final Action<DeviceFlowResponse> deviceFlowCallback) {
+    private List<Account> getUserAccounts(
+        final AtomicReference<JwtCredentials> vstsCredentialsHolder,
+        final Action<DeviceFlowResponse> deviceFlowCallback) {
         for (int retriesLeft = MAX_CREDENTIALS_RETRIES; retriesLeft > 0; retriesLeft--) {
             final Credentials vstsCredentials =
```
AlexRukhlin: Use JwtCredentials type here

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/wizard/common/WizardServerSelectionPage.java
```diff
@@ -140,6 +146,10 @@ protected boolean onPageFinished() {
                 break;
             }
 
+            if (vstsCredentials instanceof JwtCredentials) {
```
AlexRukhlin: No need to check

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/wizard/common/WizardServerSelectionPage.java
```diff
@@ -140,6 +146,10 @@ protected boolean onPageFinished() {
...
+                vstsCredentialsHolder.set((JwtCredentials) vstsCredentials);
```
AlexRukhlin: No need to cast

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/wizard/common/WizardServerSelectionPage.java
```diff
@@ -240,7 +252,7 @@ private Credentials getVstsRootCredentials(
 
         if (EnvironmentVariables.getBoolean(EnvironmentVariables.USE_OAUTH_LIBRARY, true)) {
 
-            vstsCredentials = CredentialsHelper.getOAuthCredentials(null, deviceFlowCallback);
+            vstsCredentials = CredentialsHelper.getOAuthCredentials(null, null, deviceFlowCallback);
```
AlexRukhlin: Cast to JwtCredentials here

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/wizard/common/WizardServerSelectionPage.java
```diff
@@ -240,7 +252,7 @@ private Credentials getVstsRootCredentials(
...
 
             if (vstsCredentials != null && (vstsCredentials instanceof JwtCredentials)) {
                 return vstsCredentials;
```
AlexRukhlin: Cast to JwtCredentials here

source/com.microsoft.tfs.client.common/src/com/microsoft/tfs/client/common/credentials/CredentialsHelper.java
```diff
@@ -114,11 +119,15 @@ public static Credentials getOAuthCredentials(final URI serverURI, final Action<
                 authenticator.getUriToKeyConversion().convert(serverURI, authenticator.getAuthType());
             removeStalePersonalAccessToken(tokenKey, serverURI);
 
+            final TokenPair oauth2Token =
+                (accessToken == null) ? null : new TokenPair(accessToken.getAccessToken(), "null"); //$NON-NLS-1$
```
AlexRukhlin: Why "null" but not StringUtil.Empty?

yacaovsnc: This is because the constructor for `TokenPair` doesn't take empty or null strings.  It must have a value.

PR #164, opened 2017-02-05
Update the Azure Toolkit for Eclipse link

rcjsuen:
Replace outdated link with the current link for the Azure Toolkit for
Eclipse page.


DavidStaheli:
Thanks @rcjsuen!

PR #173, opened 2017-02-27
Use UIClientConnectionAdvisor to support proxy environment

yacaovsnc:
Use UIClientConnectionAdvisor instead of CommonClientConnectionAdvisor to support proxy environment.


AlexRukhlin:
LGTM

PR #190, opened 2017-04-30
AAU I'd like to be able explicitly specifying PAT credentials for a TFS server

AlexRukhlin:
#985181


yacaovsnc:
LGTM.

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/messages.properties
```diff
@@ -1503,6 +1503,13 @@ SelectMergeVersionWizardPage.VersionTypeLabelText=&Version type:
 SelectQueryItemDialog.PromptLabelTextFormat={0}:
 SelectQueryItemDialog.SelectQueryDialogTitle=Select Query
 SelectQueryItemDialog.SelectQueryFolderDialogTitle=Select Query Folder
+ServerAuthDialog.CredentialsTypeLabel=Credentials type:
+ServerAuthDialog.DialogTitleText=Server authentication
```
DavidStaheli: "Server authentication" - both words should be capitalized in a dialog title.

source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/dialogs/connect/ServerAuthDialog.java
```diff
@@ -0,0 +1,161 @@
...
+
+    private void onCredentialTypeSelected() {
+        if (patButton.getSelection()) {
+            savedUserName = usernameText.getText();
+            usernameText.setText(StringUtil.EMPTY);
+        } else {
+            usernameText.setText(savedUserName);
+        }
+        usernameText.setEnabled(userpasswordButton.getSelection());
+    }
```
DavidStaheli: Some users may be confused if they can enter a username with a personal access token.  If they select PAT as the credential type, can we keep the username field disabled or hide it and give it some default value?  Maybe you're already doing something like this and I just need to see the UI.

PR #195, opened 2017-05-12
Bug #791785: EGit does not support Kerberos/NTLM authentication via HTTPS

AlexRukhlin:


AlexRukhlin:
A minor change in CLC is not related to the main change.

PR #204, opened 2017-06-05
Re-use location info among all REST clients

AlexRukhlin:


source/com.microsoft.tfs.core/src/com/microsoft/tfs/core/TFSConnection.java
```diff
@@ -1351,11 +1353,7 @@ public Version getServerApiVersion() {
 
         synchronized (serverApiVersionLock) {
             if (serverApiVersion == null) {
-
-                final TeeClientHandler clientHandler = new TeeClientHandler(getHTTPClient());
-                clientHandler.init(true, null, getBaseURI());
-
-                ApiResourceLocationCollection locations = clientHandler.getLocations();
+                ApiResourceLocationCollection locations = getServerApiLocations();
```
lkillgore: There's at least two locks taken when inside this line.  Do we have a convention (e.g. should we add a comment or something) to make sure that no future modifications take these two locks in a different order? (serverApiVersionLock and serverApiLocationsLock)

(Alternatively, I don't see a problem with locking on the same [serverApiVersionLock] when setting serverApiLocations.)

PR #207, opened 2017-06-08
Version 14.119.1

AlexRukhlin:
Authentication changes


source/com.microsoft.tfs.client.common.ui/src/com/microsoft/tfs/client/common/ui/config/UITransportRequestHandler.java
```diff
@@ -278,18 +277,37 @@ public Status handleException(
...
+                }
+                if (header.getValue().equals("Federated")) { //$NON-NLS-1$
+                    break;
+                }
+            }
+
             if (EnvironmentVariables.getBoolean(EnvironmentVariables.USE_OAUTH_LIBRARY, true)
-                && isPatCredentials(connectionInstanceData.getCredentials())) {
+                && usedCredentials != null
+                && new CachedCredentials(serverUrl, usedCredentials).isPatCredentials()
```
lkillgore: Would it be better to swap this line with the line below so there's no needless object creation?

AlexRukhlin: Done, thanks.

PR #212, opened 2017-06-14
Try to resolve the issue with JetBrains computer names in TFS workspaces

AlexRukhlin:


source/com.microsoft.tfs.client.clc/src/com/microsoft/tfs/client/clc/commands/Command.java
```diff
@@ -914,6 +915,58 @@ protected final WorkspaceInfo determineCachedWorkspace(final String[] pathFreeAr
...
+            log.info(e.getMessage(), e);
+        }
+
+        final String jbComputerName = LocalHost.getShortNameJB();
+        log.info(MessageFormat.format(
+            "We'll try to find the workspace one more time using the JetBrains name {0}", //$NON-NLS-1$
+            jbComputerName));
+
+        System.setProperty(LocalHost.SHORT_NAME_OVERRIDE_PROPERTY, jbComputerName);
+        return determineCachedWorkspaceImpl(pathFreeArguments, ignoreWorkspaceOptionValue);
```
DavidStaheli: If this final attempt using the JetBrains computer name is unsuccessful, should we set the system property back to its original value?

source/com.microsoft.tfs.client.clc/src/com/microsoft/tfs/client/clc/commands/Command.java
```diff
@@ -914,6 +915,58 @@ protected final WorkspaceInfo determineCachedWorkspace(final String[] pathFreeAr
...
+                    workspaceInfo.getDisplayName(),
+                    LocalHost.getShortName()));
+
+        } catch (final CannotFindWorkspaceException e) {
+            log.info(e.getMessage(), e);
+        }
+
+        final String jbComputerName = LocalHost.getShortNameJB();
+        log.info(MessageFormat.format(
+            "We'll try to find the workspace one more time using the JetBrains name {0}", //$NON-NLS-1$
```
DavidStaheli: P999: "JetBrains name" would be clearer as: "JetBrains computer name"

source/com.microsoft.tfs.client.clc/src/com/microsoft/tfs/client/clc/commands/Command.java
```diff
@@ -914,6 +915,58 @@ protected final WorkspaceInfo determineCachedWorkspace(final String[] pathFreeAr
...
+         * On the Mac OSX we should check two versions of the short computer
+         * name to ensure compatibility with the JetBrains' method of its
+         * compilation. The code should be removed when and if this
+         * compatibility is not needed anymore.
+         */
+        WorkspaceInfo workspaceInfo = null;
+        try {
+            workspaceInfo = determineCachedWorkspaceImpl(pathFreeArguments, ignoreWorkspaceOptionValue);
+
+            if (LocalHost.getShortName().equalsIgnoreCase(workspaceInfo.getComputer())) {
```
jpricketMSFT: I don't understand the need for the comparison here. If the workspace was found, shouldn't we just return it?

jpricket (dismissed):
The changes seem good to me. I had one question, but I trust that the code is correct.
Leah should probably also review them.

source/com.microsoft.tfs.client.clc/src/com/microsoft/tfs/client/clc/commands/Command.java
AlexRukhlin: IMO, we do not check the computer when trying to find a workspace in the cache. If we've found a workspace with an unexpected computer name, I would change the name (by using the system property) and continue with that. I hope that would protect us from unexpected cache refresh, which might remove the workspace.

source/com.microsoft.tfs.jni/src/com/microsoft/tfs/jni/helpers/LocalHost.java
```diff
@@ -108,10 +110,77 @@ public synchronized static String getShortName() {
                 name.substring(0, (name.length() > MAX_COMPUTER_NAME_SIZE) ? MAX_COMPUTER_NAME_SIZE : name.length());
         }
 
+        log.info("Short name resolved to: " + computerNameJB); //$NON-NLS-1$
```
leantk: Should this be computerName not computerNameJB?

source/com.microsoft.tfs.jni/src/com/microsoft/tfs/jni/helpers/LocalHost.java
```diff
@@ -108,10 +110,77 @@ public synchronized static String getShortName() {
...
+     * </p>
+     * <p>
+     * The {@link #SHORT_NAME_OVERRIDE_PROPERTY} may be used to define the
+     * computer name exactly.
+     * </p>
+     *
+     * @return a short identifier (name of this computer). Never
+     *         <code>null</code> or empty.
+     */
+    public synchronized static String getShortNameJB() {
```
leantk: Use the same way to figure out the short name like JetBrains
https://github.com/JetBrains/tfsIntegration/blob/0f127c1b10acd3f985920246f43dfbe170d69960/src/org/jetbrains/tfsIntegration/core/tfs/Workstation.java#L234

leantk (approved):
I verified this worked on my Mac

leantk:
I' seeing an issue with the workspaces command though. Investigating now...

PR #241, opened 2017-12-01
Added additional verification for disposal of credentials

andrebakker:
Previously, if some exception happened while in the for loop of
HTTPMethodDirector.executeMethod, it could occasionally exit without
releasing previously acquired handles from NTLM and Negotiate. This
change ensures the releasing of the handles by calling a cleanup method
on the finally, outside of the for loop.


andrebakker:
Closed this pull request because they had build issues. A solution is still needed but before this is stable I'll leave it closed.

PR #250, opened 2018-01-29
Ensure disposal of native auth state after executing Http Method.

brandonjacob:
Depending on the selected authorization scheme, it's possible for
some native OS handles to leak once the HttpMethodDirector has
finished executing an HttpMethod. Leaks have been observed for
Kerberos (via SPNEGO - Negotiate) but also seem possible using NTLM
authentication. Examples of potentially leaked handles include:
  * credentials handles (e.g. missing call to FreeCredentialsHandle)
  * security context (e.g. missing call to DeleteSecurityContext)

The HttpMethodDirector now ensures that the selected authorization
scheme has a chance to cleanup outstanding native OS state once the
HttpMethod has finished execution.

### DIFF
--- a/source/com.microsoft.tfs.core.httpclient/src/com/microsoft/tfs/core/httpclient/HttpMethodDirector.java
+++ b/source/com.microsoft.tfs.core.httpclient/src/com/microsoft/tfs/core/httpclient/HttpMethodDirector.java
@@ -224,6 +224,8 @@ class HttpMethodDirector {
 
             } // end of retry loop
         } finally {
+            cleanupHostAndProxyAuthState(method);
+            
             if (conn != null) {
                 conn.setLocked(false);
             }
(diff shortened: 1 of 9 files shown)


brandonjacob:
@kaylangan : I would really appreciate any feedback that you and your team can provide on this pull request.  I'm sure your team has far more expertise with the TEE codebase than I.

I have confirmed via manual testing that this patch ensures the auth state associated with an Http Method is cleaned up once the HttpMethodDirector.executeMethod completes.  

It would be nice to include some automated unit tests for this scenario.  I briefly considered doing this but quickly realized that unit testing HttpMethodDirector functionality would be non-trivial due to all of its dependencies.  Looks like a number of mocks would be necessary.  I decided to wait until I received some feedback from your team first.  If your team feels this patch is reasonable, please offer advice on the most appropriate testing strategy.

Issue #237, opened 2017-10-10
JDK 9  - No output on stdout/stderr


spoorcc:
With following versions:

- MacOS 10.7
- Tee-clc 14.123.1
- Java 1.9.181

When running `./tf` with any command, there is no output whatsoever in the terminal.
When downgrading to java 1.8.144 everything works as expected.

The command `java -version` does return output

rcnicke:
Trying again:
I had the same problem, has to do with Java 9.
Problem and solution described here: [TEE CLC not working on Mac](http://expicon.se/2017/10/13/tee-clc-not-working-on-mac/)

spoorcc:
Thanks, that might solve it. I'm pressed for time, so I downgraded to Java 8 for now.
This could then be fixed in [clc-launcher](https://github.com/Microsoft/team-explorer-everywhere/blob/c1012bc8619e265dcb6ce4c755d98df17064e5bc/source/com.microsoft.tfs.client.clc/scripts/clc-launcher) if somebody wants to make the world a better place by adding `--add-modules java.xml.bind` to the `exec java` call.

axelgenus:
This issue affects also the Windows version and the proposed solution linked by rcnicke works also on Windows.

My environment:
- Windows 10 Enterprise (64 bit)
- Java Development Kit 9.0.1+11 (64 bit)
- TEE-CLC 14.123.1

EricTheRed-VT:
I have the same env as @axelgenus , but I can't get @rcnicke 's fix to work.
I added the flag 
`java –-add-modules java.xml.bind -Xmx2048M -cp...` 
but command prompt is throwing an error: 
`Error: Could not find or load main class GÇô-add-modules                                                                Caused by: java.lang.ClassNotFoundException: GÇô-add-modules `

YasserMojahed:
I have upgraded my Android Studio to version 3 then I faced the same issue with TF.
first I added this "--add-modules java.xml.bind" to "tf"
but it still gave me the same exception.
I added it to "tf.cmd" then it worked.
I hope that wont has bad effects on "commit changes", "show diff" and other TFS - android studio functions

eric-milles:
`java.xml.bind` dependency removed in release 14.135.0

PR #251, opened 2018-02-07
Add jvm argument --add-modules java.xml.bind when java version is 9

benrobot:
Add jvm argument --add-modules java.xml.bind when java version is 9
Fix: #237

### DIFF
--- a/source/com.microsoft.tfs.client.clc/scripts/clc-launcher
+++ b/source/com.microsoft.tfs.client.clc/scripts/clc-launcher
@@ -144,6 +144,12 @@ if [ `uname` = "Linux" ] ; then
     RANDOM_DEVICE_PROPERTY="-Djava.security.egd=file:/dev/urandom"
 fi
 
-exec java -Xmx2048M -classpath "$CLC_CLASSPATH" $TF_ADDITIONAL_JAVA_ARGS \
+# Java 9 requires an additional --add-modules argument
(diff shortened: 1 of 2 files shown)


mfilippov:
@benrobot Are you hardcoded only version 9? The latest stable Java is 11.

OttoG:
While this workaround may be useful for Java 9 and 10, the java.xml.bind API (JAXB) was unfortunately removed from Java 11. See the https://stackoverflow.com/a/43574427/2416627 answer to a similar discussion.

I found that under Java 12, a sufficient workaround was to download the current version of the jaxb-api artifact from Maven as a JAR:

https://search.maven.org/search?q=g:javax.xml.bind%20AND%20a:jaxb-api

If the downloaded file (currently jaxb-api-2.4.0-b180830.0359.jar) is placed in the lib directory for TEE, the tf command starts working under JDK 12.

The proper fix should be to include this JAR as a distributed dependency library in the compiled TEE distributions.

ForNeVeR:
Another workaround is to define an environment variable `TF_NOTELEMETRY=TRUE`. Only the telemetry functions require java.xml.bind here, so no telemetry = no problems.

brunoborges:
The dependency (JAR) for XML can be downloaded and shipped with the tool, and set in the classpath as usual. No need to use modules.

PR #309, opened 2019-11-21
Fix native tests location

ForNeVeR:
This will allow them to run on the CI every time (beforehand they were set up for manual run only).


ForNeVeR:
Sorry, wrong repo.

PR #355, opened 2023-04-22
GitHub Actions+Gradle

ForNeVeR:
Let's see if this approach will work.


ForNeVeR:
Sorry, wrong target repo again.

